PR #629, opened 2026-04-22
feat(velesql): implement window functions ROW_NUMBER, RANK, DENSE_RANK (#386)

### LINKED ISSUE #386: feat(velesql): window functions (ROW_NUMBER, RANK, DENSE_RANK OVER PARTITION BY)
## Motivation

Window functions are arguably the **single most impactful missing SQL feature** for AI/RAG workloads. The core use case is "top K per group" queries:

```sql
-- Top 3 most relevant documents per source
SELECT *, ROW_NUMBER() OVER (PARTITION BY source ORDER BY similarity() DESC) AS rn
FROM documents
WHERE vector NEAR $query LIMIT 100
HAVING rn <= 3

-- Rank results with deduplication awareness
SELECT *, DENSE_RANK() OVER (PARTITION BY category ORDER BY similarity() DESC) AS rank
FROM documents
WHERE vector NEAR $query LIMIT 50
```

### Use cases for AI workloads

| Function | Use case |
|----------|----------|
| `ROW_NUMBER()` | Top-K per category in RAG retrieval |
| `RANK()` / `DENSE_RANK()` | Deduplication-aware ranking across sources |
| `LAG()` / `LEAD()` | Temporal/sequential analysis in agent episodic memory |
| `NTILE()` | Bucketing results by relevance tiers |
| `SUM() OVER` | Running similarity score totals |

### No competitor offers this

None of the major vector databases (Qdrant, Pinecone, Weaviate, Milvus, ChromaDB) offer window functions in their query languages. This would be a genuine differentiator for VelesQL.

## Work required

- Grammar: add `window_expr`, `over_clause`, `partition_by`, window function names
- AST: add `WindowFunction`, `OverClause`, `PartitionBy` types
- Parser: parse window expressions in SELECT column list
- Executor: implement window function evaluation (partition → sort → number/rank)
- Aggregator: extend `StreamingAggregator` or add separate `WindowEvaluator`

## Complexity

High — this is a significant feature. Consider phasing:
1. Phase 1: `ROW_NUMBER()` and `RANK()` with `PARTITION BY` + `ORDER BY`
2. Phase 2: `LAG`, `LEAD`, `NTILE`
3. Phase 3: Window frame (`ROWS BETWEEN ...`)

SBALAVIGNESH123:
## Summary

Implements SQL window functions `ROW_NUMBER()`, `RANK()`, `DENSE_RANK()` with `PARTITION BY` and `ORDER BY` support in VelesQL.

Closes #386

### Grammar (`grammar.pest`)
- 11 new PEG rules: `window_item`, `window_function_call`, `window_function_name`, `over_clause`, `partition_by_clause`, `partition_by_list`, `window_order_by_clause`, `window_order_by_item`, `window_order_by_expr`
- Window rules placed **before** `aggregation_item` to prevent PEG ambiguity

### AST (`ast/window.rs`) - NEW
- `WindowFunctionType` enum (`RowNumber`, `Rank`, `DenseRank`)
- `WindowOrderBy`, `OverClause`, `WindowFunction` structs
- All types derive `Serialize`/`Deserialize` with `serde(default)` for backward compatibility

### Parser (`clause_projection.rs`)
- Extended `SelectItemAccumulator` with `window_functions` tracking
- 4 new parsing methods: `parse_window_item`, `parse_window_function_call`, `parse_over_clause`, `parse_window_order_by_item`

### Executor (`window_evaluator.rs`) - NEW
- Partition-sort-rank pipeline with `BTreeMap` for deterministic ordering
- Handles nested field paths, `similarity`/`score` sorting, NULL partitions
- Integrated into `select_dispatch.rs` postprocessing (after DISTINCT, before ORDER BY)
- Injects ranking values into result payload JSON

### Tests (`window_function_tests.rs`) - NEW
- **22 tests**: 8 parser + 14 evaluator
- Edge cases: empty sets, single result, all-tied values, NULL partitions, nested fields, DESC ordering, case-insensitive keywords, similarity score sorting

## Verification

| Check | Result |
|---|---|
| `cargo check` (core, server, tauri) | All 3 crates compile |
| Clippy (new code) | Zero warnings |
| Window function tests | 22/22 pass |
| Parser regression tests | 18/18 pass |
| Validation regression tests | 75/75 pass |
| TODOs/FIXMEs/hallucinations | None found |

## Example Usage

```sql
-- Top-2 documents per source by similarity
SELECT name, source,
       ROW_NUMBER() OVER (PARTITION BY source ORDER BY similarity() DESC) AS rn
FROM documents
WHERE vector NEAR 
LIMIT 100

-- Rank with tie handling
SELECT title,
       RANK() OVER (ORDER BY score DESC) AS rnk,
       DENSE_RANK() OVER (ORDER BY score DESC) AS drnk
FROM results
```

---

<a href="https://app.devin.ai/review/cyberlife-coder/velesdb/pull/629" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,73 @@ Zero regression cumulatively —
 recall gate (Fast 0.90 / Balanced 0.95 / Accurate 0.99 / Perfect 1.00)
 passes on every phase.
 
+### Added — VelesQL window functions
+
+- **`ROW_NUMBER()`, `RANK()`, `DENSE_RANK()` with `OVER (PARTITION BY … ORDER BY …)`**
+  (`#629`, closes `#386`). Credit @SBALAVIGNESH123 for the original
+  implementation; the landed form includes a complete zero-tech-debt
+  hardening pass.
+
+  Grammar additions in `velesql/grammar.pest`: `window_item`,
+  `window_function_call`, `over_clause`, `partition_by_clause`,
+  `window_order_by_clause`. The evaluator lives in
+  `velesql/window_evaluator.rs` and runs after DISTINCT and before
+  ORDER BY / LIMIT in the query pipeline (see the design note on
+  `apply_select_postprocessing` for why the order deviates from the SQL
+  standard — vector-search survivors get a dense `1..N` numbering).
+
+  Implementation highlights (all pinned by regression tests):
+
+  - **Global snapshot-first evaluation** prevents three classes of
+    payload corruption — intra-function (alias collides with own
+    ORDER BY column), inter-function via ORDER BY, inter-function via
+    PARTITION BY. `evaluate` pre-captures every window function's
+    ORDER BY values and PARTITION BY keys **before** any injection,
+    so sequential window evaluation cannot read another function's
+    injected ranks.
+  - **Dispatch-by-variant** for rank computation (`compute_row_numbers`,
+    `compute_rank`, `compute_dense_rank`) with a shared `is_new_group`
+    tie-detection predicate — no dead state.
+  - **Canonical-JSON partition keys** preserve JSON type
+    discriminators, eliminating collisions between `Number(1)` and
+    `String("1")` and between `Null` and a literal payload string
+    `"__null__"`.
+
+### Changed — correctness fixes with visible effects
+
+These corrections resolve pre-existing bugs that were silently wrong.
+Each is pinned by regression tests; callers who depended on the buggy
+output must update.
+
+- **`SelectColumns::to_display_names` returns every SELECT-list item.**
+  Previous releases silently dropped `similarity()` expressions and
+  qualified wildcards from the Mixed SELECT variant, shortening the
+  list returned by the Python `VelesQL.parse(...).columns` getter and
+  the WASM `parsed.columns` getter. v1.13.0 returns the complete list
+  in grammar order (columns → aggregations → similarity scores →
+  qualified wildcards → window functions).
+
+  *Migration*: callers that hard-coded an expected list length or
+  `columns[n]` offsets against the buggy output will now see
+  additional entries. Update the expected length / offsets, or
+  switch to dict-style lookup by name. The complete contract is
+  pinned by
+  `velesql::ast_tests::test_display_names_mixed_includes_all_variants`.
+
+- **`DISTINCT` dedup key now includes qualified-wildcard-expanded
+  fields.** `SELECT DISTINCT ctx.*, title FROM docs` previously
+  deduped by `title` only, silently collapsing rows that differed
+  only on a wildcard-expanded field. v1.13.0 deduplicates by the full
+  payload when any qualified wildcard is in the SELECT list, matching
+  SQL's "DISTINCT considers the whole projected row" semantics.
+
+  *Migration*: queries of the shape
+  `SELECT DISTINCT alias.*, col, …` may return more rows than before.
+  If the pre-v1.13.0 behavior is required, replace `alias.*` with the
+  specific columns that should participate in the dedup key. Pinned
+  by
+  `collection::search::query::distinct_tests::tests::test_distinct_mixed_with_qualified_wildcard_dedupes_by_full_payload`.
+
 ### Added — GPU acceleration
 
 - **GPU HNSW layer-0 traversal** (`#626`, closes `#502`): SONG paper
--- a/crates/velesdb-core/src/collection/search/query/distinct.rs
+++ b/crates/velesdb-core/src/collection/search/query/distinct.rs
@@ -198,6 +198,7 @@ mod tests {
                 alias: Some("score".to_string()),
             }],
             qualified_wildcards: vec![],
+            window_functions: vec![],
         };
 
         let distinct = apply_distinct(results, &columns);
@@ -220,6 +221,7 @@ mod tests {
             aggregations: vec![],
             similarity_scores: vec![],
             qualified_wildcards: vec![],
+            window_functions: vec![],
         };
 
         let distinct = apply_distinct(results, &columns);
--- a/crates/velesdb-core/src/collection/search/query/distinct.rs
+++ b/crates/velesdb-core/src/collection/search/query/distinct.rs
@@ -9,20 +9,55 @@ use rustc_hash::FxHashSet;
 /// Apply DISTINCT deduplication to results based on selected columns (EPIC-052 US-001).
 ///
 /// Uses HashSet for O(n) complexity and preserves insertion order.
+///
+/// # Dedup-key contract
+///
+/// The SELECT-list variants drive the dedup key as follows:
+///
+/// - `Columns(cols)` → dedup by the listed payload fields only.
+/// - `Mixed { columns, qualified_wildcards, similarity_scores, .. }`:
+///   - if `qualified_wildcards` is **empty**, dedup by the listed `columns`;
+///   - if `qualified_wildcards` is **non-empty**, dedup by the **full payload**
+///     (same semantics as `SELECT *`) because a qualified wildcard expands
+///     to every payload field — those fields are part of the projected row
+///     and must participate in the dedup key, otherwise
+///     `SELECT DISTINCT ctx.*, title FROM docs` would collapse rows that
+///     differ only by non-title wildcard fields.
+///   - `similarity_scores` presence appends the similarity score to the key
+///     so rows differing only by score are not collapsed.
+///   - `aggregations` are handled by a separate aggregation pipeline and
+///     are not part of the DISTINCT key (they don't reach this path).
+///   - `window_functions` are deliberately excluded — DISTINCT runs BEFORE
+///     window evaluation per the VelesQL pipeline order (see the contract
+///     on `apply_select_postprocessing`), so any window-injected field
+///     doesn't exist on the payload yet.
+/// - `SimilarityScore(_)` → dedup by score only.
+/// - `All` / `QualifiedWildcard(_)` → dedup by full payload.
+/// - `Aggregations(_)` → no dedup (aggregations collapse rows themselves).
 pub fn apply_distinct(results: Vec<SearchResult>, columns: &SelectColumns) -> Vec<SearchResult> {
-    // If SELECT *, deduplicate by all payload fields
     let (column_names, include_score) = match columns {
         SelectColumns::Columns(cols) => (cols.iter().map(|c| c.name.clone()).collect(), false),
         SelectColumns::Mixed {
             columns: cols,
             similarity_scores,
+            qualified_wildcards,
             ..
-        } => (
-            cols.iter().map(|c| c.name.clone()).collect(),
-            !similarity_scores.is_empty(),
-        ),
+        } => {
+            // Qualified wildcards expand to every payload field; fall back
+            // to "dedup by full payload" (empty column list) so those
+            // fields participate in the key.
+            let cols_for_dedup = if qualified_wildcards.is_empty() {
+                cols.iter().map(|c| c.name.clone()).collect()
+            } else {
+                Vec::new()
+            };
+            (cols_for_dedup, !similarity_scores.is_empty())
+        }
         SelectColumns::SimilarityScore(_) => (Vec::new(), true),
-        // All, Aggregations, QualifiedWildcard: use full payload or no dedup
+        // All / QualifiedWildcard → full payload; Aggregations → no dedup
+        // (the empty-column path below gives the same "full payload" key,
+        // which is harmless for Aggregations since that path never reaches
+        // this function in practice).
         SelectColumns::All
         | SelectColumns::Aggregations(_)
         | SelectColumns::QualifiedWildcard(_) => (Vec::new(), false),
--- a/crates/velesdb-core/src/collection/search/query/distinct_tests.rs
+++ b/crates/velesdb-core/src/collection/search/query/distinct_tests.rs
@@ -101,4 +101,86 @@ mod tests {
         // Number 42 vs string "42" should be distinct (different JSON repr).
         assert_eq!(distinct.len(), 2, "number vs string should be distinct");
     }
+
+    // -----------------------------------------------------------------------
+    // Mixed with qualified_wildcards must dedup by FULL payload, not just
+    // the explicit `columns` list. Regression for the Devin finding:
+    // `SELECT DISTINCT ctx.*, title FROM docs` must collapse rows only when
+    // ALL payload fields match, not only when `title` matches.
+    // -----------------------------------------------------------------------
+
+    #[test]
+    fn test_distinct_mixed_with_qualified_wildcard_dedupes_by_full_payload() {
+        // Two rows share `title` but differ in a wildcard-expanded field
+        // (`author`). Without the fix, dedup would collapse them (only
+        // `title` considered). With the fix, dedup uses the full payload,
+        // so both survive.
+        let results = vec![
+            make_result(
+                1,
+                Some(serde_json::json!({"title": "T1", "author": "Alice"})),
+            ),
+            make_result(2, Some(serde_json::json!({"title": "T1", "author": "Bob"}))),
+            // Third row is an exact duplicate of row 1 → should collapse.
+            make_result(
+                3,
+                Some(serde_json::json!({"title": "T1", "author": "Alice"})),
+            ),
+        ];
+
+        let columns = SelectColumns::Mixed {
+            columns: vec![Column {
+                name: "title".to_string(),
+                alias: None,
+            }],
+            aggregations: Vec::new(),
+            similarity_scores: Vec::new(),
+            qualified_wildcards: vec!["ctx".to_string()],
+            window_functions: Vec::new(),
+        };
+
+        let distinct = apply_distinct(results, &columns);
+        assert_eq!(
+            distinct.len(),
+            2,
+            "rows 1 and 2 differ (Alice vs Bob) so both survive; row 3 is an \
+             exact duplicate of row 1 and is dropped"
+        );
+        // Preserves insertion order: row 1 and row 2 survive.
+        assert_eq!(distinct[0].point.id, 1);
+        assert_eq!(distinct[1].point.id, 2);
+    }
+
+    /// Control case: same query shape but WITHOUT qualified_wildcards —
+    /// dedup must still use only the explicit `columns` list (current
+    /// behaviour preserved for backward compatibility).
+    #[test]
+    fn test_distinct_mixed_without_qualified_wildcard_dedupes_by_columns_only() {
+        let results = vec![
+            make_result(
+                1,
+                Some(serde_json::json!({"title": "T1", "author": "Alice"})),
+            ),
+            make_result(2, Some(serde_json::json!({"title": "T1", "author": "Bob"}))),
+        ];
+
+        let columns = SelectColumns::Mixed {
+            columns: vec![Column {
+                name: "title".to_string(),
+                alias: None,
+            }],
+            aggregations: Vec::new(),
+            similarity_scores: Vec::new(),
+            qualified_wildcards: Vec::new(),
+            window_functions: Vec::new(),
+        };
+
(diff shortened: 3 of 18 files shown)


SBALAVIGNESH123:
Re: DISTINCT before window functions — intentional design choice

Thanks for flagging this, Devin. This is a deliberate design decision documented in both the evaluator module doc (window_evaluator.rs:6-7) and the dispatcher comments (select_dispatch.rs:267-272).

Rationale: In VelesQL's vector-search context, DISTINCT operates on the raw payload before ranking assignment. This ensures that:

Duplicate documents are collapsed first, so window functions rank only unique results — which is the expected behavior for "top-k per group" queries over a vector index.
ROW_NUMBER produces a clean 1..N sequence without gaps from removed duplicates, which is what users want for pagination and ranking use cases.
The SQL standard order (window → DISTINCT) would make DISTINCT a no-op after ROW_NUMBER (since every row gets a unique number), which defeats the purpose of deduplication in search results.

This matches VelesDB's existing pipeline where DISTINCT already runs before ORDER BY/LIMIT, and window functions slot in between these two stages consistently.

Happy to add a doc note to the grammar or user-facing docs clarifying this if helpful for the maintainer.

cyberlife-coder (changes requested):
Thanks @SBALAVIGNESH123 for this contribution — ROW_NUMBER / RANK / DENSE_RANK is a real gap in VelesQL and the overall shape of the work (grammar rules, AST node, evaluator, projection wiring, 593 lines of tests) is solid.

Unfortunately, the Devin review surfaced **three correctness bugs that block merge**. All three are real (I reproduced the reasoning locally against the diff). The existing tests don't catch them because they happen to align payload `score` field values with search `similarity` scores, which hides the interaction bugs.

## 🔴 Blocking bug 1 — `project_mixed` silently drops window function values

**Where**: `crates/velesdb-core/src/collection/search/query/projection.rs:89-126`

`SelectColumns::Mixed` is destructured with `window_functions` bound to `..`, so the selector never reads the alias back out of the payload. Result: `SELECT name, ROW_NUMBER() OVER (...) AS rn FROM products` returns `{"name": "…"}` without the `rn` field. This is user-facing — every REST response and every Python / TS SDK call that uses `SELECT … window_fn` over the API path loses the ranking column.

**Fix sketch**: in `project_single`, pass `window_functions` through to `project_mixed`; in `project_mixed`, add a loop that extracts `payload[alias_or_default_alias]` for each `WindowFunction` and inserts it into the output map, symmetric to the named-columns loop at lines 110-113.

**Test to add**: REST/SDK-level integration test asserting the output JSON contains the alias key, not just the Rust-level evaluator test.

## 🔴 Blocking bug 2 — `extract_sort_value` treats payload field `"score"` as search similarity

**Where**: `crates/velesdb-core/src/velesql/window_evaluator.rs:125`

The special-case `if column == \"similarity\" || column == \"score\"` is wrong for `"score"`. `"similarity"` is the synthetic alias the parser emits for the `similarity()` pseudo-function — legitimately maps to `result.score`. `"score"` however is a real column name the user can legitimately have in their payload (`products.score`, `documents.score`, etc.), and `ORDER BY score DESC` inside an `OVER` clause should read that payload field, not the search score.

Compare with the main ORDER BY path in `ordering.rs:385-392` which only special-cases `"fused_score"` and `"similarity"`. Window ORDER BY must behave identically for consistency — any two `ORDER BY` contexts that resolve the same column name should hit the same value.

**Fix**: drop the `|| column == \"score\"` branch. Let it fall through to the payload-field lookup.

**Test to add**: query where the payload has a `score` field with values deliberately **different** from `result.score` (e.g. `result.score ∈ [0.1..0.5]` but `payload.score ∈ [100, 90, 80]`), and assert the window function orders by the payload values.

## 🔴 Blocking bug 3 — `inject_ranking` corrupts payload during tie detection

**Where**: `crates/velesdb-core/src/velesql/window_evaluator.rs:189-219` (`assign_rankings`)

The loop:
1. Calls `inject_ranking` at line 218 to write the ranking integer into `result.payload[alias]`.
2. On the next iteration, calls `is_new_ranking_group` at line 200, which calls `extract_sort_value` on the *previous* result's payload.

When `alias` matches an ORDER BY column name (explicit alias **or** default alias), step 1 overwrites the value step 2 needs to read. Example: `RANK() OVER (ORDER BY score DESC) AS score` with ties `[100, 90, 90, 80]` produces `[1, 2, 3, 4]` instead of `[1, 2, 2, 4]`, because after position 0 gets `score: 1` injected, the pos-1 vs pos-0 comparison reads `90 vs 1` (corrupted, not a tie) and then pos-2 vs pos-1 reads `90 vs 2` (still not a tie). Default aliases (e.g. `RANK() OVER (ORDER BY rank DESC)` with a payload field called `rank`) hit the same trap.

**Fix approach** (per Devin, and the right structural call): split the loop in two phases.

```rust
// Phase 1 — compute all (index, ranking) pairs reading live payload.
let mut rankings: Vec<(usize, u64)> = Vec::with_capacity(results.len());
let mut current_rank = 1;
// ... existing tie logic, but pushing into rankings instead of mutating
// results[i].payload directly ...

// Phase 2 — commit all rankings in one pass, no more reads after writes.
for (i, rank) in rankings {
    inject_ranking(&mut results[i], rank, alias);
}
```

**Test to add**: the tie-detection failure case above, with deliberately-colliding `ORDER BY score AS score` and `ORDER BY col1 AS col1` covering both explicit and default alias paths. These are the tests that would have caught the bug at authoring time.

## 🟡 Warning-level findings

Four additional 🟡 / 🚩 items are worth addressing in the same rework pass — see the full [Devin review](https://app.devin.ai/review/cyberlife-coder/velesdb/pull/629). They're not blocking individually but the rework on the critical path is likely to touch the same code, so batching makes sense.

## CI approval

Your CI hasn't fully run yet because you're a first-time contributor — only Codacy ran. Push the fixes above and I'll approve the Actions runs on the new SHA so Lint / Tests / Windows / WASM / etc. all execute. That way we validate the fixes end-to-end before merge.

## Follow-up once green

Once the three blocking bugs are fixed, tests added, and the full CI matrix is green, ping me and I'll give the green light. This isn't making v1.13.0 (releasing now for pre-seed milestone), so we have time to get window functions right for v1.14.

Thanks again for the patience and the contribution — the feature itself is wanted. Just needs the three corrections above before it can ship.

cyberlife-coder:
Update — I took your snapshot-based fix (commit 20cb6229) as the baseline, rebased on current \`develop\` (so the CI disk-space fix from #644 is in scope), resolved the 4 remaining \`clippy::pedantic\` errors that surface under the \`--all-targets\` CI sweep, and added three regression tests — one per original critical finding — so neither of the three bugs can silently come back.

**Clippy fixes**
- \`parse_window_item_parts\` tuple → named \`WindowItemParts\` type alias
- \`evaluate()\` gets an \`# Errors\` doc section
- \`extract_nested_value\` uses \`let Some(payload) = … else\`
- \`window_function_tests::get_payload_u64\` uses method reference \`serde_json::Value::as_u64\`

**Regression tests added**
- \`test_order_by_payload_score_is_not_hijacked_by_similarity\` — payload \`score\` diverges from search similarity, ordering must use payload value (bug #2).
- \`test_rank_alias_collides_with_order_by_column_preserves_ties\` — \`RANK() OVER (ORDER BY score DESC) AS score\` with \`[100, 90, 90, 80]\` must emit \`[1, 2, 2, 4]\` (bug #3, explicit alias).
- \`test_rank_default_alias_collides_with_payload_field_preserves_ties\` — same pattern via default alias path (bug #3, default alias).

**Verification (CI-mirror, local)**
- \`cargo clippy --workspace --all-targets --features persistence,gpu,update-check --exclude velesdb-python -- -D warnings -D clippy::pedantic\` → 0 errors
- \`cargo fmt --all --check\` → clean
- \`cargo check -p velesdb-core --no-default-features\` → clean
- 25 window-function tests pass (22 existing + 3 new).

Force-pushed to your branch via maintainer-can-modify. Approving the workflow runs now so the full CI matrix validates end-to-end. When CI is green I'll flip my review to ✅ and merge.

Thanks again for the original work and the snapshot refactor — that's the correct structural fix for bug #3.

cyberlife-coder:
## Review status — all 🔴 critical findings resolved

Taking a pass over the four 🔴 findings in the latest Devin report. Three were already fixed in this branch but Devin hadn't posted a ✅ Resolved marker. The fourth was a real new bug and is fixed in the commit I just pushed (`b67f65ea`).

| # | Finding | Status | Where it is fixed |
|---|---------|--------|-------------------|
| 1 | `project_mixed` silently drops window functions | ✅ Already fixed | `collection/search/query/projection.rs:100-141` — `project_mixed` now accepts and iterates `window_functions` |
| 2 | `extract_sort_value` hijacks payload `score` | ✅ Already fixed | `velesql/window_evaluator.rs:274-279` — only `"similarity"` is special-cased |
| 3 | Intra-function alias-collision corrupts own sort key | ✅ Already fixed | `velesql/window_evaluator.rs:107-156` — three-phase snapshot → compute → inject per function |
| 4 | **Inter-function contamination across sequential window functions** | ✅ **Fixed in `b67f65ea`** | `velesql/window_evaluator.rs:86-104` — snapshot hoisted to Phase 0 of `evaluate`, captured for **every** window function before ANY injection |

### About finding #4 (the new bug)

Devin is correct that the prior code only protected *within* a single window function. Two window functions with a colliding alias/column pair still corrupted each other:

```sql
SELECT ROW_NUMBER() OVER (ORDER BY score DESC) AS score,
       RANK()       OVER (ORDER BY score DESC) AS rnk
FROM t
```

On `[100, 90, 90, 80]` this produced `rnk = [1, 2, 3, 4]` instead of `[1, 2, 2, 4]` — ties were destroyed because RANK's snapshot was taken *after* ROW_NUMBER had overwritten `payload["score"]` with `[1, 2, 3, 4]`.

**Fix shape**: `evaluate()` now pre-captures `WindowSnapshot { order_by_values, partition_keys }` for every window function *before* calling `apply_single_window` for any of them. Injection is deferred to the per-function Phase 3 as before.

Two regression tests cover the fix:
- `test_inter_function_contamination_via_order_by_column` — exact scenario above
- `test_inter_function_contamination_via_partition_by_column` — same mechanism but via `PARTITION BY` collision (4 original-group rows would have become 4 singleton partitions after the first injection)

### Other findings

- **`partition_key` separator**: already uses `\x1F` (ASCII Unit Separator) — see `window_evaluator.rs:186`. The finding referenced `|` but the code has not used `|` at any point in this branch; probably stale from an earlier snapshot of the file.

### Verification

- All 4708 `velesdb-core` lib tests pass (`--test-threads=1`).
- `cargo clippy --all-targets --features persistence -- -D warnings -D clippy::pedantic` is clean.
- `cargo fmt --all` applied.

Re-requesting Devin review.

cyberlife-coder:
## Zero tech-debt pass — 5 🚩 findings resolved (commit `d2573f6a`)

After the 4 🔴 criticals were closed, I took a pass on every 🚩 informational finding from the last Devin review so no tech debt remains. Five are addressed by this commit, the rest are either intentional, pre-existing out-of-scope, or confirmed non-bugs.

### Fixed

| # | Finding | Fix |
|---|---------|-----|
| 1 | Partition key: `Number(1)` vs `String(\"1\")` collision | `extract_payload_value` now uses canonical JSON (`value.to_string()`), preserving type discriminators. Consistent with `distinct::compute_distinct_key`. |
| 2 | Partition key: `Null` vs literal `\"__null__\"` string collision | Same fix — `null` literal vs quoted `\"__null__\"` are now distinct. |
| 3 | `to_display_names` omits `similarity_scores` + `qualified_wildcards` in Mixed | Every SELECT-list variant contributes a display name, in grammar order: columns → aggregations → similarity_scores → qualified_wildcards → window_functions. |
| 4 | `project_mixed` re-inserts window values already expanded by wildcards | Pre-computed `FxHashSet<&str>` of window aliases, skip them during wildcard expansion. Single source of truth for the window branch. |
| 5 | `compute_rankings` dead state (`rank` init when DenseRank, vice-versa) | Dispatch to `compute_row_numbers` / `compute_rank` / `compute_dense_rank`; each maintains only its own state. Shared `is_new_group` predicate factored out. |

### Documented as intentional design (not debt)

| Finding | Why it stays |
|---------|--------------|
| DISTINCT runs before window functions (deviates from SQL standard) | Vector-search contract: `SELECT DISTINCT title, ROW_NUMBER() ... ` must number *deduped survivors* with a contiguous `1..N`, not leave gaps after DISTINCT drops rows. Now documented explicitly in the doc comment on `apply_select_postprocessing` with the full rationale + escape hatch for a future SQL-compat mode. Cross-linked from `window_evaluator` module header. Pinned by `test_row_number_numbers_deduped_rows_contiguously`. |
| `inject_ranking` may overwrite an existing payload field with the same name as the alias | Aligned with LET bindings' behaviour (`select_dispatch.rs:327-330`). This is the documented semantics; a runtime warning would spam per-row. Users control the alias. |
| `extract_sort_value` special-cases `\"similarity\"` | Consistent with the main ORDER BY system. A payload field literally named `similarity` is a naming conflict the user controls. |
| `grammar.pest` ordering of RANK/DENSE_RANK | Devin confirmed in this review: \"correct since they share no prefix\". |
| Window-only SELECT falls through to `Mixed` variant with empty columns | Devin confirmed \"intentional and correct\". |

### Pre-existing, out of PR scope

| Finding | Rationale |
|---------|-----------|
| Partition key with Object/Array values | These shouldn't appear in partition columns in practice; canonical JSON renders them consistently anyway. |
| Existing `distinct.rs` uses the same `\x1F` separator | Not a new concern — already the codebase convention. |

### New regression tests

- `test_partition_key_does_not_collide_int_and_string_of_same_digits`
- `test_partition_key_distinguishes_null_from_literal_sentinel`
- `test_row_number_numbers_deduped_rows_contiguously`
- `test_display_names_all_wildcard`, `_columns_only`, `_similarity_score_with_alias`, `_similarity_score_no_alias`, `_qualified_wildcard`, `_mixed_includes_all_variants`, `_mixed_without_columns_still_exposes_score_and_wildcard` (7 in `ast_tests.rs`)

### Verification

- 4708+ `velesdb-core` lib tests pass (`--test-threads=1`); the one transient HNSW recall flake re-ran green.
- `cargo clippy --all-targets --features persistence -- -D warnings -D clippy::pedantic` clean.
- `cargo fmt --all` applied.

Ready for Devin re-review.

cyberlife-coder:
## Follow-up on the 2 new 🚩 findings — commit `e07d59ee`

### Finding 1 — DISTINCT dedup key ignores `qualified_wildcards`

Devin called this "pre-existing, not introduced by this PR", but under a zero-tech-debt policy pre-existing debt we touch is still debt we own. **Fixed.**

`apply_distinct` now falls back to the full-payload key whenever `Mixed.qualified_wildcards` is non-empty, so `SELECT DISTINCT ctx.*, title FROM docs` deduplicates on every wildcard-expanded field — not just on `title`. The full dedup-key contract for every `SelectColumns` variant is now documented on the function. Two regression tests in `distinct_tests.rs` pin the behaviour:

- `test_distinct_mixed_with_qualified_wildcard_dedupes_by_full_payload` — rows identical only on `title` but differing on `author` now survive (they'd collapse pre-fix).
- `test_distinct_mixed_without_qualified_wildcard_dedupes_by_columns_only` — control case, confirms explicit-columns behaviour is preserved.

### Finding 2 — `to_display_names` behavioural change visible to bindings

Devin is correct that my earlier fix is a behavioural change — but it's the **correctness** fix for a bug Devin itself flagged two reviews ago (silent drop of `similarity_scores` + `qualified_wildcards`). **Documented explicitly.**

Added a dedicated `# Completeness` section on `to_display_names` explaining:
- the historical incomplete behaviour was a bug,
- the new contract returns exactly one entry per SELECT-list item,
- callers who hard-coded `len()` against the shorter list must update,
- the new contract is pinned by `ast_tests::test_display_names_mixed_includes_all_variants`.

No code change needed beyond the doc note; the behaviour change is desired.

### Status

- Gates: `cargo clippy --all-targets -- -D warnings -D clippy::pedantic` clean, `cargo fmt --all` applied.
- Tests: 4719+ lib tests pass (the one transient HNSW recall flake re-ran green — unrelated to VelesQL).
- No 🔴 findings open. No 🚩 findings open.

Re-requesting Devin review.

cyberlife-coder:
## CTO-grade pass: zero debt + zero user frustration (commit `b1c814b7`)

Both Devin findings were technically bug fixes, but shipping a fix silently still creates frustration for anyone whose pipeline was reading the old output. Here is what ships to close that gap without adding debt:

### 1. CHANGELOG v1.13.0 — explicit migration notes

Added a complete window-functions entry under `Added` (crediting @SBALAVIGNESH123, linking the design, listing the regression tests), and a new `Changed — correctness fixes with visible effects` subsection covering:

- **`to_display_names` contract completion** — explains the pre-fix list silently dropped `similarity()` and qualified wildcards, shows the migration (update expected length / switch to name-based lookup), links the pinning test.
- **DISTINCT + qualified wildcards** — explains that `SELECT DISTINCT alias.*, col, …` may now return more rows, shows how to replicate pre-v1.13.0 behavior (replace the wildcard with explicit columns), links the pinning test.

### 2. Python + WASM API docstrings

`VelesQL.parse(...).columns` now carries a dedicated *"v1.13.0 contract completion"* note in:
- The PyO3 docstring (surfaced via `help(parsed.columns)`).
- The WASM JSDoc (surfaced in the generated TypeScript type exports).

Users reading the docs see exactly what changed and what to update — no silent surprise.

### 3. End-to-end BDD coverage — `tests/bdd/window_functions.rs` (8 tests)

Not just evaluator-level, but full pipeline from SQL string through `Database::execute_query`:

| Test | What it pins |
|------|--------------|
| `test_row_number_over_order_by_score_desc` | ROW_NUMBER injects a dense `rn` into every row's payload |
| `test_rank_partition_by_source_order_by_score_desc` | PARTITION BY rankings stay independent across partitions |
| `test_dense_rank_single_partition_with_ties_has_no_gaps` | DENSE_RANK semantics (`1, 2, 2, 3`) end-to-end |
| `test_distinct_then_window_numbers_survivors_contiguously` | VelesQL-intentional DISTINCT-before-window order → contiguous `1..N` |
| `test_distinct_with_qualified_wildcard_dedupes_by_full_payload` | The pre-existing DISTINCT + wildcard bug is fixed end-to-end |
| `test_rank_alias_collides_with_payload_score_end_to_end` | Intra-function corruption fix holds through the full pipeline |
| `test_window_function_on_empty_result_returns_no_rows` | Negative path: empty filter does not panic |
| `test_window_function_coexists_with_vector_search_near` | Window slot runs on NEAR-ranked results |

### Gate status

- 388 BDD tests pass (all previously-passing + 8 new).
- `cargo clippy --all-targets --features persistence -- -D warnings -D clippy::pedantic` clean.
- `cargo fmt --all` applied.

**Net**: every 🚩 finding from the last three Devin passes is now either fixed in code or documented as an explicit, tested design choice. Users see every behavior change before they trip on it. No debt, no frustration.

cyberlife-coder (approved):
Approved after the zero-tech-debt + user-frustration pass.

All 🔴 critical findings closed (inter-function contamination fix + ✅ Resolved marker from Devin), all 🚩 informational findings either fixed in code or documented as explicit design choices with pinning tests.

CI fully green — Codacy, Devin, Snyk + all four previously-blocked fork workflows (CI, PR Governance, Propagation Guard, Compatibility Matrix).

Credit @SBALAVIGNESH123 for the original implementation.

SBALAVIGNESH123:
Thanks for merging @cyberlife-coder — really enjoyed working on this! The rigorous review process and your extra zero-tech-debt fixes on top made the feature incredibly solid.

I'd love to keep contributing to VelesDB . Quick question — are there any job or contract opportunities on your team right now? I'm actively looking and would love to bring this kind of Rust/systems work into a full-time role. Either way, happy to keep contributing to the project!

PR #626, opened 2026-04-21
perf(gpu): GPU-accelerated HNSW layer-0 traversal (SONG 3-stage pipeline) — rebased from #622

cyberlife-coder:
## Summary

Implements GPU-based BFS/DFS traversal for HNSW layer 0 via wgpu compute shaders, following the [SONG paper](https://arxiv.org/abs/2012.01012).

**Supersedes and replaces [#622](https://github.com/cyberlife-coder/VelesDB/pull/622)** — original work by @SBALAVIGNESH123, cherry-picked onto a fresh branch from `develop` because #622's branch had no common merge-base with `develop` (structural divergence, freshness gate unfixable via rebase/merge).

**Closes #502**

All 11 original commits are preserved with original authorship (Bala Vignesh S as author on 10 of them, plus my u32 correctness gate and dead_code cleanup).

## Architecture

```
Query -> CPU greedy descent (layers max->1) -> GPU SONG pipeline (layer 0) -> Results
```

### SONG 3-Stage Pipeline (all in 1 CommandBuffer)

| Stage | Shader | Purpose |
|-------|--------|---------|
| Expand | `EXPAND_FRONTIER_SHADER` | Parallel BFS via CSR with atomic visited bitset |
| Distance | `TRAVERSAL_{EUCLIDEAN_SQ,COSINE,DOT_PRODUCT}_SHADER` | Batch distance for all new candidates |
| Select | `SELECT_TOPK_SHADER` | Bitonic sort in workgroup-local memory |

## Key Design Decisions

- Activation threshold: GPU only for `>500K vectors` AND `num_vectors * dim <= u32::MAX` (correctness gate — WGSL has no u64)
- Single command buffer: all iterations in 1 encoder (no per-iter CPU-GPU sync)
- Double-buffered frontier: ping-pong swap
- Adaptive iterations: 10-20 based on `ef_search`
- Graceful fallback: returns `None` on any GPU error → CPU SIMD path
- Generation-based CSR cache invalidation on insert/parallel_insert

## Diff vs #622

Zero behavioural difference from the latest state of #622 (SHA `0d2810df`), just:

- Re-based onto current `develop` (includes PR template Git Flow notice from #624)
- Dead-code warnings on `search_auto` / `search_gpu` silenced with explicit reasons
- Unused `CsrGraph` import removed from `gpu_search.rs`

## Security review

[...]

PR #637, opened 2026-04-22
chore(gpu): enforce lock-ordering for gpu_vectors_snapshot + CsrCache caller contract

### LINKED ISSUE #634: GPU traversal — post-merge hardening (5 PRs covering 8 Devin findings from PR #626)
## Context

[PR #626](https://github.com/cyberlife-coder/VelesDB/pull/626) landed the SONG 3-stage GPU HNSW traversal. Devin review produced a mix of 🔴 critical (all fixed in-PR), 🚩 ANALYSIS flags, and performance observations. This issue tracks **every non-blocker finding** worth addressing post-merge, organised into 5 bundled PRs.

> **Revised after a full re-read of the Devin transcript** — the initial version of this issue covered only 3 items. Five additional items surfaced that bundle cleanly with the first three or open new fronts.

---

## 🚨 Priority 0 — `search_auto` is never called in production

**File**: [crates/velesdb-core/src/index/hnsw/native_inner.rs:202-218](https://github.com/cyberlife-coder/VelesDB/blob/develop/crates/velesdb-core/src/index/hnsw/native_inner.rs#L202-L218)

The `search_auto` entry point that dispatches between CPU and GPU paths is annotated `#[allow(dead_code)]` and has **zero callers**. The main search pipeline still calls `search()` directly, which means every line of GPU code added by #626 is exercised only by unit tests and benchmarks — never in production query flow.

Devin's words:
> GPU search path is effectively untested in production flow — only unit tests and benchmarks exercise it.

This isn't tech debt, it's **shipped-but-dormant**. Fixing the three lock-ordering findings without first wiring `search_auto` in is re-arranging deck chairs on a feature nobody uses.

### Plan — PR-D (half-day, moderate risk, **highest priority**)

- [ ] Identify the canonical callers of `NativeHnsw::search` in the search pipeline (search handlers, server dispatch, Python bindings).
- [ ] Route them through `search_auto` when the `gpu` feature is enabled.
- [ ] Add a feature-gated integration test that forces the GPU path and asserts result parity with CPU on a small (≤ 100K) index.
- [ ] Remove the `#[allow(dead_code)]` or replace with a narrower attribute.
- [ ] Recall gate at 1M scale (cargo test `test_recall` already in the suite).
- [ ] Drop the stale "search_auto never called" note from native_inner.

**Expected gain**: the GPU feature becomes reachable. Makes every subsequent optimisation (PR-C, PR-E) meaningful rather than theoretical.

---

## 🟢 Priority 1 — Lock ordering `gpu_vectors_snapshot` / `vectors` + CsrCache doc contract

**Finding 1 — Implicit lock ordering**
**File**: [crates/velesdb-core/src/index/hnsw/native/graph/gpu_search.rs:170-191](https://github.com/cyberlife-coder/VelesDB/blob/develop/crates/velesdb-core/src/index/hnsw/native/graph/gpu_search.rs#L170-L191)

`get_or_refresh_vector_snapshot` holds `gpu_vectors_snapshot` (Mutex) while calling `with_vectors_read`. `insert.rs:110-114` acquires the opposite order but releases `vectors.write()` first. No ABBA today, but there is no `LockRank` entry for `gpu_vectors_snapshot` in `graph/locking.rs` — a future change that nests these locks differently could silently deadlock.

**Finding E — `CsrCache::get_or_rebuild` "CAS" comment is misleading**
**File**: [crates/velesdb-core/src/gpu/gpu_csr.rs:296-334](https://github.com/cyberlife-coder/VelesDB/blob/develop/crates/velesdb-core/src/gpu/gpu_csr.rs#L296-L334)

Comments claim a "CAS on generation" but the code is a simple load-compare-store sequence. Today it is safe only because **all callers hold the layers read lock** for the entire rebuild, so overlapping rebuilds see identical layer state. If a future caller ever invokes `get_or_rebuild` without the layers lock, stale CSRs could overwrite fresh ones without tripping the generation check.

### Plan — PR-A (~1 h, zero risk)

- [ ] Add `LockRank::GpuVectorsSnapshot` in `crates/velesdb-core/src/index/hnsw/native/graph/locking.rs` between existing ranks enclosing the pattern (verify against current rank values).
- [ ] Instrument the two acquisition sites (`gpu_search.rs:174-175` and `insert.rs:110-114`) with the new rank.
- [ ] Fix the misleading comment on `CsrCache::get_or_rebuild` — describe the actual mechanism (check-then-write under write lock, invariant that callers hold layers read lock).
- [ ] Add `debug_assert!` inside `get_or_rebuild` that a layers read lock is effectively held (or make the function take a witness type that proves it, such as `&LayersReadGuard`).
- [ ] Run full test suite + concurrent integration tests.

**Expected gain**: static enforcement of both lock invariants. Two future classes of silent bug become compile / debug-assert errors.

---

## 🟡 Priority 2 — Unified version counter for snapshot AND CSR caches

**Finding 2 — Snapshot cache keyed on `count`**
**File**: [crates/velesdb-core/src/index/hnsw/native/graph/gpu_search.rs:177-179](https://github.com/cyberlife-coder/VelesDB/blob/develop/crates/velesdb-core/src/index/hnsw/native/graph/gpu_search.rs#L177-L179)

Today safe because `insert` explicitly clears the snapshot, but delete+insert returning to the same count would serve stale vectors silently. Implicit invariant — any future delete path that forgets the clear ritual breaks this.

**Finding F — CSR cache check uses `Ordering::Relaxed` count**
**File**: [crates/velesdb-core/src/index/hnsw/native/graph/gpu_search.rs:70-81](https://github.com/cyberlife-coder/VelesDB/blob/develop/crates/velesdb-core/src/index/hnsw/native/graph/gpu_search.rs#L70-L81)

Same pattern on the CSR cache: `existing.num_nodes as usize == count` with `count` loaded Relaxed. Tiny race window between load and cache check. Worst case: one query served slightly stale results.

### Plan — PR-B (~2 h, low risk)

- [ ] Add `snapshot_version: AtomicU64` on `NativeHnsw` that bumps on **any** vectors / layers mutation (insert ✅, delete when added, any graph-structure change).
- [ ] Snapshot cache stores `(version_at_build, dim, Arc<[f32]>)`; valid iff `current_version == stored_version`.
- [ ] `CsrCache` already has a generation counter — align it with the same versioning scheme so one bump invalidates both caches.
- [ ] Keep existing explicit `None`-clears as belt-and-suspenders.
- [ ] Regression test: simulate a delete path that forgets to clear — cache must still detect staleness via version.

**Expected gain**: one mutation bumps one counter, both caches invalidate correctly. No future mutator needs to remember a ritual.

---

## 🟠 Priority 3 — Measure before optimising: concurrent-insert stall

**Finding 3 — Snapshot mutex held during `vectors.read()` and snapshot build**
**File**: [crates/velesdb-core/src/index/hnsw/native/graph/gpu_search.rs:174-191](https://github.com/cyberlife-coder/VelesDB/blob/develop/crates/velesdb-core/src/index/hnsw/native/graph/gpu_search.rs#L174-L191)

Mutex held while copying up to ~1.5 GB of flat vectors. No deadlock; concurrent inserters stall for the snapshot build duration.

**Finding G — Layers read lock held during full CSR rebuild (~50 ms at 1 M nodes)**
**File**: [crates/velesdb-core/src/index/hnsw/native/graph/gpu_search.rs:70-81](https://github.com/cyberlife-coder/VelesDB/blob/develop/crates/velesdb-core/src/index/hnsw/native/graph/gpu_search.rs#L70-L81)

The CSR rebuild runs inside `with_layers_read`, blocking all writes (inserts) for the rebuild duration. CSR cache amortises it but the first GPU search after a batch of inserts pays the full cost.

### Plan — PR-C (half-day, moderate risk, **measure first**)

- [ ] Criterion bench `concurrent_insert_vs_gpu_search` with configurable thread counts, 1M-vector graph, invalidation cycles. Measures **both** stalls simultaneously (snapshot build + CSR rebuild).
- [ ] If combined stall < 5 % of insert throughput → **do nothing**, add a comment pointing at this bench so the next person can re-measure.
- [ ] If stall significant:
  - [ ] Snapshot refresh: double-checked locking — check cache under mutex (O(1)), if stale drop mutex, compute snapshot, re-acquire + CAS install. Race acceptable (two rebuilders → last wins, other's work discarded).
  - [ ] CSR rebuild: consider releasing layers read lock after snapshotting the layer adjacency into an `Arc<Layer>`, rebuilding the CSR from the snapshot outside the lock.
  - [ ] Loom model verification.

**Expected gain**: only if benchmarks justify it. Otherwise: explicit "intentionally simple, measured" comment.

---

## 🔵 Priority 4 — Parallel `select_topk` after perf bench

**Finding I — `select_topk` shader uses single-thread insertion sort**
**File**: [crates/velesdb-core/src/gpu/shaders.rs:428-487](https://github.com/cyberlife-coder/VelesDB/blob/develop/crates/velesdb-core/src/gpu/shaders.rs#L428-L487)

`@workgroup_size(1)` doing an O(n × k) insertion sort over 8 192 candidates per iteration, × 10-20 iterations per query. Devin labels it as **likely the bottleneck of the GPU pipeline**. The single-thread design is intentional (the parallel design was incorrect), but correct-and-slow is not the final state.

### Plan — PR-E (1-2 days, moderate risk)

- [ ] **Prerequisite**: PR-D merged (no point optimising a path nobody calls) + PR-C bench exists to measure delta.
- [ ] Prototype a correct parallel reduction: per-workgroup local sort of 256 candidates → k-way merge of the 32 local top-k lists in a second pass.
- [ ] Verify recall parity with the serial implementation using the existing recall gate.
- [ ] Compare GPU latency vs current single-thread implementation at 1 M / 5 M scales.
- [ ] Keep the serial shader as a fallback / golden reference for correctness tests.

**Expected gain**: removes the intentional perf ceiling. Only worth doing once the GPU path is live (PR-D) and measured (PR-C).

---

## Scheduling

| PR | Description | Effort | Risk | Prereq |
|----|-------------|--------|------|--------|
| **D** | Wire `search_auto` in production pipeline | ½ day | Medium | — |
| **A** | LockRank + CsrCache doc fix + debug_assert | 1 h | Zero | — |
| **B** | Unified version counter (snapshot + CSR) | 2 h | Low | — |
| **C** | Bench + optional stall fixes | ½ day | Medium | A, B |
| **E** | Parallel `select_topk` | 1-2 days | Medium | D, C |

A, B, D are independent and can land in parallel. C requires A/B to have version counter semantics in place; E requires D to make the work actually relevant and C to provide before/after numbers.

## Anti-patterns to avoid

- ❌ Bundling all five into one "GPU hardening" PR — different risk profiles kill reviewability.
- ❌ Adding `// TODO(...)` in the code — repo rules forbid orphan TODOs. This issue IS the tracker.
- ❌ Optimising C or E without a bench.
- ❌ Skipping D and "coming back to it later" — the rest of this issue is cosmetic if the feature is never reached in production.

## Findings explicitly confirmed safe / no-action

For reference, the following Devin findings on PR #626 were verified safe or were already resolved in-PR. No action needed:

- `gpu_backend.rs:6` — visibility `pub(super) → pub(crate)` is correct and minimal.
- `gpu_csr.rs:277` — `is_stale()` two-load TOCTOU is benign; fast path is best-effort, slow path is guarded by write lock.
- `gpu_search.rs:71` (first) — lock ordering `layers → gpu_vectors_snapshot → vectors` verified: `layers.read()` is released before `gpu_vectors_snapshot.lock()`, no actual nested violation.
- `gpu_traversal.rs:118` — u32 overflow gate correctly prevents silent GPU distance corruption at > u32::MAX offsets.
- `insert.rs:109` — GPU cache invalidation correctly added to single + batch insert paths + `file_load`.
- `gpu_traversal.rs:777` — `candidate_dists` not reset between iterations; safe because the sentinel pattern in `candidates` + distance shader overwrite every slot each iteration.
- `gpu_buffer_cache.rs` — already removed in commit 026f976f; the orphan-module finding no longer applies.

cyberlife-coder:
## Summary

Addresses the first two 🚩 ANALYSIS findings from Devin on [PR #626](https://github.com/cyberlife-coder/VelesDB/pull/626), tracked in [issue #634 as PR-A](https://github.com/cyberlife-coder/VelesDB/issues/634).

No behaviour change in production — pure invariant instrumentation + doc contract tightening.

Closes #634 for PR-A.

### Finding 1 — `gpu_vectors_snapshot` lock ordering was implicit

Devin:
> Lock ordering between gpu_vectors_snapshot and vectors is safe but fragile ... the ordering is implicit — there's no LockRank entry for gpu_vectors_snapshot in the locking framework.

Fix:
- Add `LockRank::GpuVectorsSnapshot = 5` (strictly lower than `Vectors = 10`).
- Instrument the three acquisition sites with `record_lock_acquire` / `record_lock_release`:
  - `gpu_search.rs::get_or_refresh_vector_snapshot`
  - `insert.rs` (post-vectors-write invalidation)
  - `backend_adapter.rs::parallel_insert` (same invalidation)
- Debug builds now catch ordering violations via the existing thread-local stack; release builds keep the F-25 zero-overhead path.
- Module-level doc in `locking.rs` updated to show the new rank in the ordering diagram.

### Finding E — `CsrCache::get_or_rebuild` "CAS on generation" was misleading

Devin:
> The doc comment at line 326 claims "CAS on generation" protects against this, but no CAS is used — it's a load-compare-store sequence with the cache write already committed. ... if `get_or_rebuild` were ever called WITHOUT holding the layers read lock, a thread could write a stale CSR after another thread writes a fresh one.

Fix:
- Rewrite the doc to describe the actual mechanism (check-then-write under `self.csr.write()` guarded by `gen_before`/`gen_after`).
- Document the caller contract explicitly: must hold `LockRank::Layers`.
- Add a `debug_assert!(holds_lock(Layers))` tripwire via a new crate-internal `holds_lock` helper in `locking.rs`.
- New `hnsw_holds_lock` / `HnswLockRank` re-exports through `native/mod.rs` keep `mod graph` private while letting `crate::gpu::gpu_csr` assert the contract.

### Tests

- `locking.rs`: 3 new unit tests cover `holds_lock` behaviour, rank monotonicity, nested acquire/release round-trips.
- `gpu_csr.rs::tests`: existing cache tests wrapped in a `with_layers_rank` helper that models the production caller contract (`with_layers_read → get_or_rebuild`).
- `gpu_csr_tests.rs`: same helper applied to the 4 concurrent/regression tests + new `#[should_panic(expected = "layers read lock")]` test guards the contract itself.

## Verification (CI-mirror, local)

| Check | Result |
|---|---|
| `cargo clippy --workspace --all-targets --features persistence,gpu,update-check --exclude velesdb-python -- -D warnings -D clippy::pedantic` | ✅ 0 errors |
| `cargo fmt --all --check` | ✅ |
| `cargo check -p velesdb-core --no-default-features` | ✅ |
| `cargo test -p velesdb-core --features gpu --lib gpu::` | ✅ 79 pass, 0 fail |
| `cargo test -p velesdb-core --features persistence,gpu --test recall_validation` | ✅ 7 pass, 1 ignored (manual) |

## Out of scope for this PR

Tracked in #634:
- PR-B (version counter unifying snapshot + CSR invalidation) — next up
- PR-C (bench concurrent-insert stall, then optional non-blocking refresh)
- PR-D (wire `search_auto` into production pipeline) — running in parallel right now
- PR-E (parallel `select_topk` after bench)

Issue #634 is updated with the full plan.

---

<a href="https://app.devin.ai/review/cyberlife-coder/velesdb/pull/637" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/crates/velesdb-core/src/gpu/gpu_csr.rs
+++ b/crates/velesdb-core/src/gpu/gpu_csr.rs
@@ -296,12 +296,46 @@ impl CsrCache {
     ///
     /// This method is designed for the GPU dispatch hot path:
     /// - If the cache is fresh, returns the existing CSR (fast path).
-    /// - If stale, rebuilds from the Layer (slow path, ~50ms for 1M nodes).
+    /// - If stale, rebuilds from the `layer` argument (slow path, ~50 ms
+    ///   for 1 M nodes).
     ///
-    /// The rebuild is generation-safe: if a concurrent `invalidate()` occurs
-    /// during the rebuild, the stale CSR is still stored but the generation
-    /// check ensures the *next* query will re-trigger a rebuild.
+    /// # Caller contract — must hold `layers` read lock
+    ///
+    /// **The rebuild from `layer` is only race-free while the caller holds
+    /// the `layers` read lock** (rank 20 in `locking::LockRank`) for the
+    /// entire duration of this call. Layer mutations require the write
+    /// lock, so as long as the read lock is held, every concurrent
+    /// rebuilder sees the same layer topology and produces an identical
+    /// CSR. The `gen_before`/`gen_after` check under `self.csr.write()`
+    /// then discards any rebuild that raced with an `invalidate()`.
+    ///
+    /// If a future caller ever invokes `get_or_rebuild` **without** the
+    /// layers read lock held, two rebuilders could observe different
+    /// layer states, and one might commit a stale CSR while the
+    /// generation counter indicates fresh — a silent correctness bug.
+    /// Debug builds assert the precondition via `holds_lock(Layers)`.
+    ///
+    /// # Generation protocol (NOT a CAS)
+    ///
+    /// The rebuild is **not** a compare-and-swap. It is a
+    /// load-before-rebuild (`gen_before`) / re-check-under-write-lock
+    /// (`gen_after`) sequence: we store the new CSR and bump
+    /// `built_generation` only when the generation has not moved during
+    /// the rebuild. Correctness depends on the caller contract above —
+    /// without the shared layer snapshot, the write would still be
+    /// atomic but the data committed could be stale.
     pub fn get_or_rebuild(&self, layer: &Layer, num_nodes: usize) -> CsrGraph {
+        // Caller contract: the layers read lock must be held so concurrent
+        // rebuilders observe the same layer topology. In release builds
+        // `hnsw_holds_lock` returns `true` unconditionally (thread-local
+        // stack is not maintained), so the assert is a debug-only tripwire.
+        debug_assert!(
+            crate::index::hnsw::native::hnsw_holds_lock(
+                crate::index::hnsw::native::HnswLockRank::Layers
+            ),
+            "CsrCache::get_or_rebuild must be called while holding the layers read lock"
+        );
+
         // Fast path: check if cache is fresh
         if !self.is_stale() {
             let guard = self.csr.read();
@@ -313,7 +347,9 @@ impl CsrCache {
         // Snapshot generation before rebuild
         let gen_before = self.generation.load(Ordering::Acquire);
 
-        // Slow path: rebuild outside of any lock.
+        // Slow path: rebuild outside of any lock on `self.csr`. Safe
+        // against concurrent rebuilders because they all observe the
+        // same layer state (see caller contract above).
         let new_csr = CsrGraph::from_layer(layer, num_nodes);
 
         // Commit is atomic under the write lock: we only store our CSR
@@ -368,7 +404,21 @@ impl Default for CsrCache {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::index::hnsw::native::NodeId;
+    use crate::index::hnsw::native::{
+        hnsw_holds_lock, hnsw_record_lock_acquire, hnsw_record_lock_release, HnswLockRank, NodeId,
+    };
+
+    /// Runs `f` with the layers rank recorded as held — the caller contract
+    /// of [`CsrCache::get_or_rebuild`]. Abstracts away the record/release
+    /// dance so tests read linearly and do not accidentally break the
+    /// contract by forgetting the release.
+    fn with_layers_rank<R>(f: impl FnOnce() -> R) -> R {
+        hnsw_record_lock_acquire(HnswLockRank::Layers);
+        debug_assert!(hnsw_holds_lock(HnswLockRank::Layers));
+        let result = f();
+        hnsw_record_lock_release(HnswLockRank::Layers);
+        result
+    }
 
     #[test]
     fn test_csr_from_empty_layer() {
@@ -432,19 +482,20 @@ mod tests {
         layer.set_neighbors(0, vec![1]);
         layer.set_neighbors(1, vec![0]);
 
-        // First build
-        let csr = cache.get_or_rebuild(&layer, 2);
+        // First build — tests model the production caller contract
+        // (`with_layers_read → get_or_rebuild`) via `with_layers_rank`.
+        let csr = with_layers_rank(|| cache.get_or_rebuild(&layer, 2));
         assert_eq!(csr.num_nodes, 2);
         assert_eq!(cache.version(), 1);
 
         // Should return cached (not rebuild)
-        let csr2 = cache.get_or_rebuild(&layer, 2);
+        let csr2 = with_layers_rank(|| cache.get_or_rebuild(&layer, 2));
         assert_eq!(csr2.num_nodes, 2);
         assert_eq!(cache.version(), 1); // Same version
 
         // Invalidate and rebuild
         cache.invalidate();
-        let csr3 = cache.get_or_rebuild(&layer, 2);
+        let csr3 = with_layers_rank(|| cache.get_or_rebuild(&layer, 2));
         assert_eq!(csr3.num_nodes, 2);
         assert_eq!(cache.version(), 2); // Incremented
     }
@@ -484,7 +535,7 @@ mod tests {
         assert!(cache.get_if_clean().is_none()); // Starts dirty
 
         let layer = Layer::new(1);
-        cache.get_or_rebuild(&layer, 1); // Build it
+        with_layers_rank(|| cache.get_or_rebuild(&layer, 1)); // Build it
         assert!(cache.get_if_clean().is_some()); // Now clean
 
         cache.invalidate();
--- a/crates/velesdb-core/src/gpu/gpu_csr_tests.rs
+++ b/crates/velesdb-core/src/gpu/gpu_csr_tests.rs
@@ -7,6 +7,19 @@
 
 use crate::gpu::gpu_csr::{CsrCache, CsrGraph};
 use crate::index::hnsw::native::layer::{Layer, NodeId};
+use crate::index::hnsw::native::{
+    hnsw_record_lock_acquire, hnsw_record_lock_release, HnswLockRank,
+};
+
+/// Runs `f` with the layers rank recorded as held — the caller contract
+/// of [`CsrCache::get_or_rebuild`]. Mirrors the helper in `gpu_csr::tests`
+/// so concurrent tests spread across threads keep their rank stack clean.
+fn with_layers_rank<R>(f: impl FnOnce() -> R) -> R {
+    hnsw_record_lock_acquire(HnswLockRank::Layers);
+    let result = f();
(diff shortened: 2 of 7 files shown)


cyberlife-coder:
Devin 🟡 addressed: the doc at locking.rs:142 was corrected to say "Returns `true` in release builds" in commit 8a0b4de8 (it matches the stub at line 161 and the caller doc at gpu_csr.rs:330). Pre-fix review pass showed the stale `false` wording; current tip is consistent.

PR #638, opened 2026-04-22
feat(gpu): wire search_auto into production pipeline (closes #634 PR-D)

### LINKED ISSUE #634: GPU traversal — post-merge hardening (5 PRs covering 8 Devin findings from PR #626)

cyberlife-coder:
## Summary

Wires `NativeHnswInner::search_auto` into the three production HNSW search
entry points, activating the GPU layer-0 traversal that shipped as unreachable
code in v1.13.0 (PR #626).

Before this change `search_auto` carried `#[allow(dead_code)]` with no caller
in the query pipeline — the v1.13.0 release notes advertised
"GPU-accelerated HNSW search" but production traffic never hit the GPU path.

## Change set

| Layer | Before | After |
|---|---|---|
| `HnswIndex::search_hnsw_only` (trait `search`, `search_with_quality`, batch, rerank paths) | `inner.search(q, k, ef)` | `inner.search_auto(q, k, ef)` |
| `HnswIndex::search_hnsw_only_filtered` (bitmap-pre-filtered search) | `inner.search(...)` | `inner.search_auto(...)` |
| `NativeHnswIndex::search_with_quality` (public direct-access wrapper) | `inner.search(...)` | `inner.search_auto(...)` |
| `NativeHnswInner::search_auto` / `search_gpu` | `#[allow(dead_code)]` | live, wired |

The GPU gate is entirely internal to `search_auto`:
* `#[cfg(feature = "gpu")]` — compile-time
* `should_traverse_gpu(len, dim)` — run-time: `len > 500_000 && len * dim <= u32::MAX`
* `HnswBackend::Standard` — RaBitQ stays on CPU (binary GPU shader not implemented)

Below the 500K threshold, without `gpu` feature, or on RaBitQ, the call is a
verbatim tail-call to the original CPU `search` — zero hot-path overhead.

## Impact matrix

| Crate | Impact | Action |
|---|---|---|
| `velesdb-core` | source of change | wired + tested |
| `velesdb-server` | uses `Collection::search` → `HnswIndex` → wired | cargo check ✓ |
| `velesdb-wasm` | `gpu` feature unavailable; search_auto always tail-calls search | cargo check wasm32 no-default-features ✓ |
| `velesdb-cli`, `velesdb-mobile`, `velesdb-migrate`, `tauri-plugin-velesdb` | same — route through core `HnswIndex` | cargo check ✓ |
| `velesdb-python` | N/A (excluded per project rules) | — |
| TypeScript SDK | no API change | — |

## Verification

- [x] `cargo fmt --all --check`
- [x] `cargo clippy --workspace --all-targets --features persistence,gpu,update-check --exclude velesdb-python -- -D warnings -D clippy::pedantic` — clean
- [x] `cargo clippy -p velesdb-core --features persistence --all-targets -- -D warnings -D clippy::pedantic` (no-gpu build) — clean
- [x] `cargo check -p velesdb-core --no-default-features` — compiles
- [x] `cargo check -p velesdb-wasm --target wasm32-unknown-unknown --no-default-features` — compiles
- [x] `cargo test -p velesdb-core --features persistence --lib hnsw -- --test-threads=1` — 546/546 pass
- [x] `cargo test -p velesdb-core --features persistence,gpu --lib hnsw -- --test-threads=1` — 559/559 pass
- [x] `cargo test -p velesdb-core --features persistence --test recall_validation -- --test-threads=1` — 7/7 pass (recall gate)
- [x] `cargo test -p velesdb-core --features persistence,gpu --test recall_validation -- --test-threads=1` — 7/7 pass
- [x] `cargo test -p velesdb-core --features persistence --test bdd -- --test-threads=1` — 380/380 pass
- [x] `cargo test -p velesdb-core --features persistence --lib collection -- --test-threads=1` — 1441/1441 pass

## New test coverage (`gpu_search_auto_tests.rs`)

5 tests (4 always-on + 1 feature-gated `gpu`):

| Test | Purpose |
|---|---|
| `search_auto_matches_cpu_search_below_gpu_threshold_cosine` | Bit-exact `(id, score)` parity vs direct `inner.search` for Cosine at 2K vectors |
| `search_auto_matches_cpu_search_below_gpu_threshold_euclidean` | Same parity for Euclidean (squared-L2 → L2 transform) |
| `search_auto_wiring_preserves_recall_small_dataset` | Trait-level `VectorIndex::search` still yields recall@10 ≥ 0.80 vs brute-force |
| `search_auto_wiring_preserves_score_ordering_on_accurate_quality` | Cosine scores remain monotonically non-increasing after wiring |
| `search_auto_below_threshold_matches_cpu_under_gpu_feature` | (gpu feature) Asserts sub-threshold GPU builds still tail-call CPU search verbatim |

The 500K+ GPU path itself is already exercised by `benches/gpu_traversal_benchmark.rs`.

## Risks for human review

1. **GPU behavior at 500K+**: I do not own hardware to unit-test a 500K+ index
   at CI scale. The existing bench covers it; once this PR merges, manual
   large-scale recall testing on a GPU-equipped runner is recommended.
2. **RaBitQ fallback path**: `search_auto` returns `None` from `search_gpu`
   for RaBitQ and falls through to CPU. Verified via inspection and the
   feature-gated unit test, but not exercised with a RaBitQ-mode collection
   at GPU-threshold scale.
3. **Cache coherence**: `NativeHnsw::gpu_csr_cache` has its own invalidation
   logic from PR #626 (per-instance, generation-based). This PR only adds
   callers — the cache correctness story is unchanged.

Closes #634 PR-D (GPU dead-code remediation, sub-task D only).
Does NOT touch: PR-A (locking.rs / gpu_csr.rs / gpu_vectors_snapshot) or
PR-B (version counter / CSR cache key) — those are separate sub-tasks.

---

<a href="https://app.devin.ai/review/cyberlife-coder/velesdb/pull/638" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **GPU HNSW traversal now reachable from the query pipeline** (`#634`
+  PR-D). In v1.13.0 the GPU layer-0 traversal shipped as unreachable
+  code: `NativeHnswInner::search_auto` existed but was annotated
+  `#[allow(dead_code)]` with no production caller. The three in-process
+  search entry points (`HnswIndex::search_hnsw_only`,
+  `HnswIndex::search_hnsw_only_filtered`,
+  `NativeHnswIndex::search_with_quality`) now route through
+  `search_auto`, so any build with the `gpu` feature and an index above
+  500K vectors will automatically use the GPU SONG pipeline. Sub-500K
+  indices, RaBitQ backends, and builds without `gpu` continue on the
+  CPU SIMD path with zero overhead (tail-call to the original `search`).
+
 ## [1.13.0] — 2026-04-23
 
 ### Summary
--- a/crates/velesdb-core/src/index/hnsw/gpu_search_auto_tests.rs
+++ b/crates/velesdb-core/src/index/hnsw/gpu_search_auto_tests.rs
@@ -1,0 +1,270 @@
+//! Tests for the GPU-aware `search_auto` routing in the production HNSW path.
+//!
+//! These tests verify that the CPU/GPU dispatch wiring (PR #634, PR-D) is
+//! correctly integrated at the `HnswIndex` and `NativeHnswInner` layer:
+//!
+//! 1. **Parity**: at sub-threshold sizes (below `should_traverse_gpu`'s
+//!    500K vector gate) `search_auto` falls through to the CPU path and
+//!    must yield identical `(node_id, raw_dist)` tuples as the direct
+//!    CPU `search` call.
+//!
+//! 2. **Routing**: the production call-sites (`search_hnsw_only`,
+//!    `search_hnsw_only_filtered`) still return correctly ranked results
+//!    after being switched to `search_auto`.
+//!
+//! 3. **Backend isolation**: `search_auto` on a Standard backend only
+//!    touches GPU code when gated; on `RaBitQ` it always stays on CPU.
+//!    We cannot construct a 500K+ index in a unit test to exercise the
+//!    GPU path itself — that is covered by `gpu_traversal_benchmark.rs`.
+//!
+//! The tests run without the `gpu` feature (CPU-only path) and, when the
+//! feature is enabled, additionally verify that the gated path still
+//! produces parity with the direct CPU path for sub-threshold indices.
+
+#![allow(clippy::cast_precision_loss)]
+
+use super::index::HnswIndex;
+use super::params::SearchQuality;
+use crate::distance::DistanceMetric;
+use crate::index::VectorIndex;
+
+/// Deterministic pseudo-random vector generator (no `rand` dependency).
+///
+/// Mirrors the pattern used in `gpu_rerank_tests.rs` so test data is
+/// reproducible across runs and machines.
+fn pseudo_random_vector(dim: usize, seed: &mut u64) -> Vec<f32> {
+    (0..dim)
+        .map(|_| {
+            *seed = seed.wrapping_mul(6_364_136_223_846_793_005).wrapping_add(1);
+            (*seed >> 33) as f32 / u32::MAX as f32 * 2.0 - 1.0
+        })
+        .collect()
+}
+
+/// Builds a small HNSW index populated with deterministic vectors.
+fn build_small_index(dim: usize, count: u64, metric: DistanceMetric) -> HnswIndex {
+    let index = HnswIndex::new(dim, metric).expect("test: new HNSW index");
+    let mut seed: u64 = 0x00C0_FFEE;
+    for id in 0..count {
+        let v = pseudo_random_vector(dim, &mut seed);
+        index.insert(id, &v);
+    }
+    index
+}
+
+// =========================================================================
+// Parity — search_auto falls back to CPU below the GPU threshold
+// =========================================================================
+
+/// GIVEN a small HNSW index (sub-500K vectors — below `should_traverse_gpu`)
+/// WHEN the production `search_hnsw_only` is invoked (now wired through
+///      `NativeHnswInner::search_auto`)
+/// THEN it returns the same ranked set as the direct `inner.search` call,
+///      proving the CPU fallback path inside `search_auto` is honoured.
+#[test]
+fn search_auto_matches_cpu_search_below_gpu_threshold_cosine() {
+    let dim = 64;
+    let k = 10;
+    let ef_search = 128;
+    let index = build_small_index(dim, 2000, DistanceMetric::Cosine);
+
+    let query: Vec<f32> = (0..dim).map(|j| (j as f32 * 0.0173).sin()).collect();
+
+    // Path under test — production call-site routes through search_auto.
+    let auto_results = index.search_hnsw_only(&query, k, ef_search);
+
+    // Reference — direct inner.search (pre-PR behaviour).
+    let ref_results: Vec<_> = {
+        let inner = index.inner.read();
+        let neighbours = inner.search(&query, k, ef_search);
+        neighbours
+            .into_iter()
+            .filter_map(|(node_id, raw_dist)| {
+                index.mappings.get_id(node_id).map(|id| {
+                    let score = inner.transform_score(raw_dist);
+                    crate::scored_result::ScoredResult::new(id, score)
+                })
+            })
+            .collect()
+    };
+
+    assert_eq!(
+        auto_results.len(),
+        ref_results.len(),
+        "search_auto and direct search must return the same number of results below the GPU threshold",
+    );
+    let auto_ids: Vec<u64> = auto_results.iter().map(|r| r.id).collect();
+    let ref_ids: Vec<u64> = ref_results.iter().map(|r| r.id).collect();
+    assert_eq!(
+        auto_ids, ref_ids,
+        "search_auto must return the same IDs in the same order as CPU search below the GPU threshold",
+    );
+    for (a, b) in auto_results.iter().zip(ref_results.iter()) {
+        assert!(
+            (a.score - b.score).abs() < f32::EPSILON * 16.0,
+            "scores must match exactly: auto={} vs ref={}",
+            a.score,
+            b.score,
+        );
+    }
+}
+
+/// Same parity check for Euclidean — ensures the distance transform
+/// (squared-L2 → L2) is consistent across both paths.
+#[test]
+fn search_auto_matches_cpu_search_below_gpu_threshold_euclidean() {
+    let dim = 32;
+    let k = 5;
+    let ef_search = 96;
+    let index = build_small_index(dim, 1500, DistanceMetric::Euclidean);
+
+    let query: Vec<f32> = (0..dim).map(|j| (j as f32 * 0.021).cos()).collect();
+
+    let auto_results = index.search_hnsw_only(&query, k, ef_search);
+    let ref_results: Vec<_> = {
+        let inner = index.inner.read();
+        let neighbours = inner.search(&query, k, ef_search);
+        neighbours
+            .into_iter()
+            .filter_map(|(node_id, raw_dist)| {
+                index.mappings.get_id(node_id).map(|id| {
+                    let score = inner.transform_score(raw_dist);
+                    crate::scored_result::ScoredResult::new(id, score)
+                })
+            })
+            .collect()
+    };
+
+    let auto_ids: Vec<u64> = auto_results.iter().map(|r| r.id).collect();
+    let ref_ids: Vec<u64> = ref_results.iter().map(|r| r.id).collect();
+    assert_eq!(auto_ids, ref_ids, "Euclidean parity below GPU threshold");
+}
+
+// =========================================================================
+// Routing — public search paths still produce correctly ranked results
+// =========================================================================
+
+/// GIVEN a 1000-vector index (all sub-threshold, CPU path only)
+/// WHEN we search via the public `HnswIndex::search` trait method
+///      (goes through `search_with_quality(Balanced)` → `search_hnsw_only`
+///      → `NativeHnswInner::search_auto`)
+/// THEN the top-k recall against brute force is above 0.80 — confirming
+///      the wiring did not break graph traversal correctness.
+#[test]
+fn search_auto_wiring_preserves_recall_small_dataset() {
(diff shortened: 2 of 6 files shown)


cyberlife-coder:
Devin 🚩 acknowledged: the `ef_search` scaling divergence between `HnswIndex::search_with_quality` (uses `ef_search_for_scale(k, self.len())`) and `NativeHnswIndex::search_with_quality` (uses base `ef_search(k)`) is **pre-existing** — this PR only swaps `search` → `search_auto` without touching the ef calculation. The routing PR surfaces the inconsistency but does not create it.

Opening a follow-up issue to align the two paths on `ef_search_for_scale`. Not blocking this PR since:
1. Aligning the behaviour is a semantic change (recall characteristics shift) that wants its own PR with before/after recall numbers.
2. Both paths remain functionally correct — the GPU traversal honours whatever `ef_search` it receives; the inconsistency is about which `ef` the CPU-level quality helper picks, not about GPU correctness.
3. The landing of this PR (search_auto wired in) is a release-notes promise for v1.13.0; deferring it on a pre-existing scaling nit trades a concrete delivery for a documentation problem.

Issue #379, opened 2026-03-24
feat: developer experience — simplify onboarding to <5 min


cyberlife-coder:
## Strategic Context

> DuckDB won because it was **simple**, not because it was the fastest OLAP engine.

Developer experience is THE adoption multiplier. VelesDB's technical performance is strong — what's missing is frictionless onboarding. The target: any developer goes from zero to first search result in under 5 minutes.

## Current Friction Points

1. **Too many collection types** — `Collection` (deprecated), `VectorCollection`, `GraphCollection`, `MetadataCollection`. New users don't know which to use.
2. **No zero-config mode** — Users must specify dimension, metric, storage mode. Could auto-detect from first insert.
3. **VelesQL learning curve** — Powerful but not discoverable. No interactive playground.
4. **Error messages** — Technical (VELES-xxx codes) but not always actionable for beginners.
5. **Python SDK** — `velesdb` package requires Rust toolchain via `maturin develop`. Pre-built wheels needed for all platforms.

### Quick wins (v1.8)
- [ ] Auto-detect dimension from first vector insert (no need to specify upfront)
- [ ] Default metric = Cosine (most common use case for text embeddings)
- [ ] `Database::open_or_create()` — single call for the common case
- [ ] Better error messages with suggested fixes (e.g., "dimension mismatch: expected 768, got 384 — check your embedding model")

### Medium-term (v1.9)
- [ ] VelesQL autocomplete in CLI REPL
- [ ] Pre-built Python wheels for Linux/macOS/Windows (manylinux2014, universal2)
- [ ] `velesdb init` CLI command that scaffolds a project with example code

### Long-term (v2.0)
- [ ] Unified `Collection` type that auto-routes to Vector/Graph/Metadata based on usage
- [ ] Schema inference from inserted data
- [ ] One-liner quickstart: `let db = VelesDB::quickstart("./data")?;`

## Success Metrics
- New Rust user: `cargo add` → first search result in <5 minutes
- New Python user: `pip install` → first search result in <3 minutes
- Zero questions needed to complete the getting-started guide

cyberlife-coder:
[...]

PR #709, opened 2026-04-28
fix(ts-sdk): WasmBackend.init() works in Node.js (real DX bug)

cyberlife-coder:
## Summary

`new VelesDB({ backend: 'wasm' })` followed by `db.init()` crashes 100% of the time in Node.js with:

```
Failed to initialize WASM module
  cause: Error: not implemented... yet...
    at schemeFetch (node:internal/deps/undici/undici:10891:34)
```

The wasm-pack `default()` initializer fetches the .wasm file via the global `fetch()`. Browsers handle this fine. Node.js's undici has no scheme handler for `file://` URLs, so the import explodes the first time.

The JSDoc on `VelesDB` ([sdks/typescript/src/index.ts:8](sdks/typescript/src/index.ts)) advertises *"WASM backend (browser/Node.js)"* — so this was a real DX bug, not a documented limitation.

## Discovered via

While building dx-timing onboarding scenarios for [#379](https://github.com/cyberlife-coder/VelesDB/issues/379), the Node container scenario crashed at `db.init()`. Reproducible on Node 20 + `@wiscale/velesdb-sdk@1.13.6`.

## Fix

- `WasmBackend.init()` ([wasm.ts:126](sdks/typescript/src/backends/wasm.ts)) detects Node via `process.versions.node` and, when running there, reads the `velesdb_wasm_bg.wasm` bytes from disk via `createRequire` + `node:fs/promises.readFile`, then passes them to the wasm-pack initializer as an explicit `Uint8Array`. Browsers keep the fetch path unchanged.
- `WasmModule.default()` type signature updated to accept the optional `Uint8Array | URL | string` arg the wasm-pack runtime actually supports.

## Validation

- 26/26 existing wasm-backend tests pass with the fix (no regression on browser-style mock).
- Manual smoke test in Docker `node:20-slim` container with `@wiscale/velesdb-sdk` installed locally:
  ```
  first match: id=1 score=1.0000
  ```
  Reproducible via the dx-timing scenario being prepared for #379.

## Impact

- **Browser**: zero change. The Node detection branch never executes in a browser environment (no `process` global).
- **Node.js**: the WASM backend now works. `new VelesDB({ backend: 'wasm' })` + `db.init()` succeeds.
- **REST backend**: unchanged.

## Release note

This fix should ship as v1.13.7 patch so the public npm package is usable from Node.js (not just browsers). No bump required for Rust crates / Python wheel — they are unaffected.

---

<a href="https://app.devin.ai/review/cyberlife-coder/velesdb/pull/709" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/sdks/typescript/src/backends/wasm-node-loader.ts
+++ b/sdks/typescript/src/backends/wasm-node-loader.ts
@@ -1,0 +1,83 @@
+/**
+ * Node-only helpers for the WASM backend.
+ *
+ * Browsers initialize the wasm-pack module via its default `fetch` path —
+ * no extra plumbing is needed there. Node consumers, however, fail at
+ * `default()` because Node's stdlib `fetch` has no `file://` scheme handler
+ * (the import explodes with "not implemented... yet..."). This module
+ * isolates the Node detection + bytes-loader so `wasm.ts` stays under the
+ * 500 NLOC limit (.claude/rules/code-quality.md).
+ *
+ * The functions here are referenced from `wasm.ts#init()` exactly when
+ * {@link isNodeRuntime} returns true; in browser bundles the readdir/fs/
+ * require imports are never hit because the conditional fences them off.
+ */
+
+// Ambient declaration so this TypeScript source can reference Node's CJS
+// `__filename` global without dragging in `@types/node`. The runtime check
+// `typeof __filename !== 'undefined'` keeps ESM bundles safe — the variable
+// only resolves under a CJS module wrapper.
+declare const __filename: string | undefined;
+
+/**
+ * True iff we're running under a Node.js runtime. Centralized so both the
+ * `init()` branch decision in `WasmBackend` and the bytes-loader below use
+ * the same signal.
+ */
+export function isNodeRuntime(): boolean {
+  return (
+    typeof process !== 'undefined' &&
+    Boolean((process as { versions?: { node?: string } }).versions?.node)
+  );
+}
+
+/**
+ * Read the wasm binary from disk in Node.js, returning a `Uint8Array` that
+ * the wasm-pack `default()` initializer accepts as a `BufferSource`.
+ *
+ * Implementation notes:
+ *   - The WASM filename is **discovered at runtime by listing the package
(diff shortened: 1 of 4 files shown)


cyberlife-coder:
### Code review

Found 3 issues:

1. `wasm.ts` reaches 510 lines after this PR, exceeding the per-file limit ([`.claude/rules/code-quality.md`](https://github.com/cyberlife-coder/VelesDB/blob/04e54e27985cef41ac0cd8ae81f8cd6964847426/.claude/rules/code-quality.md) says "File NLOC | **≤ 500** | Per source file"). Codacy enforces this gate. The new `isNodeRuntime` + `loadWasmBytesNode` helpers (~75 lines) should move to a sibling `wasm-node-loader.ts` so `wasm.ts` stays under the limit.

https://github.com/cyberlife-coder/VelesDB/blob/04e54e27985cef41ac0cd8ae81f8cd6964847426/sdks/typescript/src/backends/wasm.ts#L505-L510

2. `loadWasmBytesNode` JSDoc claims the WASM filename is *"discovered at runtime by reading `@wiscale/velesdb-wasm/package.json#files` for any `*.wasm` entry"*, but the implementation explicitly avoids that path — the body of the function reads `We deliberately do NOT inspect package.json#files` and uses `readdir` instead (commit 16e30a43). Stale doc directly contradicts the code under it; future maintainers will be misled.

https://github.com/cyberlife-coder/VelesDB/blob/04e54e27985cef41ac0cd8ae81f8cd6964847426/sdks/typescript/src/backends/wasm.ts#L449-L455

3. Same JSDoc block claims `createRequire(import.meta.url)` *"works under both module systems **without per-format branching**"*, but the implementation immediately below performs explicit `__filename` vs `import.meta.url` branching (because tsup's CJS output leaves `import.meta.url` as `undefined`). Doc and code are mutually inconsistent.

https://github.com/cyberlife-coder/VelesDB/blob/04e54e27985cef41ac0cd8ae81f8cd6964847426/sdks/typescript/src/backends/wasm.ts#L454-L460


<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #796, opened 2026-05-14
docs(agent-memory): document ACT-R decay + DiminishingReturns + CompositeStrategy (closes #433)

### LINKED ISSUE #433: Agent Memory SDK: implement ACT-R inspired procedural learning
## Context

The current `procedural.reinforce()` uses a flat linear adjustment: +0.1 on success, -0.05 on failure, clamped to [0, 1]. While simple and predictable, this does not capture how skills are actually reinforced through experience.

The [ACT-R cognitive architecture](https://www.researchgate.net/publication/329493100_ACT-R_A_cognitive_architecture_for_modeling_cognition) (Anderson, 1996) provides well-researched mechanisms for procedural learning that could make VelesDB's Agent Memory SDK significantly smarter.

### Phase 1: Base-level activation decay

Procedures that haven't been used recently should naturally lose confidence over time, following a power-law decay (the same forgetting curve observed in human memory research).

**Current behavior:** A procedure reinforced 6 months ago and never used since still has the same confidence.
**Expected behavior:** Confidence decays with `t^(-d)` where `t` is time since last use and `d` is a decay parameter (~0.5 in ACT-R).

This requires tracking `last_used_at` timestamp on each procedure (the internal `usage_count` field already exists in the binary).

### Phase 2: Context-sensitive utility learning

Instead of a fixed +0.1/-0.05, modulate confidence adjustment based on usage history:

```
delta = base_delta / (1 + k * success_count)  // diminishing returns
```

Early successes boost confidence more than later ones (learning curve). Early failures penalize more heavily (quick rejection of bad procedures).

### Phase 3: Spreading activation across memory types

When `procedural.recall()` is called, optionally boost procedures that are contextually related to recently accessed semantic or episodic memories. This would connect the three memory subsystems into a coherent cognitive model.

Example: if the agent just accessed semantic memory about "RAG pipeline" and episodic memory about "slow search", a procedure about "debug_rag_performance" should get a contextual boost even if its embedding similarity is lower.

## References

- Anderson, J.R. (1996). ACT-R: A Theory of Higher Level Cognition. Human-Computer Interaction, 12(4), 439-462.
- Anderson, J.R., & Lebiere, C. (1998). The Atomic Components of Thought. Lawrence Erlbaum Associates.
- Squire, L.R. (2004). Memory systems of the brain. Neurobiology of Learning and Memory, 82(3), 171-177.

## Notes

- Phase 1 is the most impactful and relatively simple to implement
- Phase 2 can reuse the existing internal `success_count`/`failure_count` fields
- Phase 3 is more experimental and would require API changes
- All phases should be backward-compatible (decay and spreading activation can be optional parameters)

cyberlife-coder:
## Summary
Brings `docs/guides/AGENT_MEMORY.md` in sync with the procedural-memory work that shipped in [#780](https://github.com/cyberlife-coder/VelesDB/pull/780) but never made it to the user-facing guide.

- **Reinforcement strategies table**: was 4 rows, now 6 — adds `DiminishingReturns` (`delta = base / (1 + k · count)`, Rescorla-Wagner 1972 / ACT-R Phase 2) and `CompositeStrategy` (weighted blend).
- **CompositeStrategy snippet**: 70% FixedRate + 30% TemporalDecay example.
- **New subsection** "Apply activation decay at recall time" — documents `ProceduralMemory::with_activation_decay()`, the ACT-R Phase 1 base-level power-law decay (`max(1, t_days)^(-d)`, Anderson 1996, d ≈ 0.5), the read-only-at-recall semantics, the opt-in default, and the Rust-only binding gap.

## Why
Issue #433 asked for ACT-R-inspired procedural learning in three phases. Phases 1 (decay) and 2 (Rescorla-Wagner diminishing returns) **already landed** via PR #780 (commit `cedf234e`), but the guide still listed only the four pre-existing strategies and made no mention of the decay knob — so the issue stayed open even though most of the work was done. This PR closes that loop.

Phase 3 (cross-memory spreading activation, explicitly labelled "experimental" in the original issue) is **not** addressed here — it needs its own API design pass and is scoped out for a future ticket.

## Test plan
- [x] All formulas cross-checked against \`crates/velesdb-core/src/agent/reinforcement.rs\` (\`power_law_decay\` line 537, \`DiminishingReturns::effective_delta\` line 482, \`CompositeStrategy::update_confidence\` lines 408–433) and \`procedural_memory.rs\` (\`with_activation_decay\` line 161, clamp range \`[0.0, 2.0]\`)
- [x] Citations verified — Rescorla-Wagner 1972 for DiminishingReturns (matches the code comment at \`reinforcement.rs:440\`), Anderson 1996 for activation decay
- [x] Python + TS binding gap verified via grep — \`with_activation_decay\` is not exposed
- [x] \`/simplify\` reviewed (3 agents): citation corrected (was misattributed to Anderson), heading rewritten to verb-form to match the rest of the file, redundant lead-in trimmed

Closes #433.

### DIFF
--- a/docs/guides/AGENT_MEMORY.md
+++ b/docs/guides/AGENT_MEMORY.md
@@ -525,7 +525,7 @@ Each `.vamm` file contains:
 
 ## Reinforcement Strategies
 
-Procedural memory confidence scores can be updated with 4 strategies:
+Procedural memory confidence scores can be updated with one of six strategies:
 
 | Strategy | Behavior | Best For |
 |----------|----------|----------|
@@ -533,6 +533,8 @@ Procedural memory confidence scores can be updated with 4 strategies:
 | **AdaptiveLearningRate** | Learning rate decays with usage | Stable procedures |
 | **TemporalDecay** | Confidence decays over time (30-day half-life) | Perishable knowledge |
 | **ContextualReinforcement** | Mix: 30% success rate + 30% usage + 40% recency | Sophisticated evaluation |
+| **DiminishingReturns** | `delta = base / (1 + k · count)` — early reinforcements weigh more than later ones (Rescorla-Wagner 1972; ACT-R Phase 2) | Fast convergence on novel procedures |
+| **CompositeStrategy** | Weighted average of any combination of the strategies above | Tuning a custom blend |
 
 The default strategy is `FixedRate`. Advanced strategies are available via the Rust API:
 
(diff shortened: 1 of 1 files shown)


cyberlife-coder:

PR #1082, opened 2026-06-11
fix: quality residuals — REST/SDK parity, GraphFirst anchors, PQ persistence, TTL TOCTOU, WeightedRRF, Parallel planner

cyberlife-coder:
## Summary

Addresses 9 tracked residuals from the 2026-06 quality audit across the core engine, query planner, quantization layer, and TypeScript SDK.

### EPIC-015 — Strict-schema node-type validation (`store_node_payload`)

Node payload writes (`INSERT NODE`) were previously unchecked in strict-schema graph collections — undeclared node types silently persisted. `validate_node_labels_against_schema` now mirrors the existing edge-validation pattern: in strict mode every `_labels` entry is checked against `GraphSchema` before any WAL write; schemaless collections and labelless payloads short-circuit at zero cost. Rejection is atomic — no partial state on failure.

### EPIC-040 — `FusionStrategy::WeightedRRF` with 0-based ranks

Added `FusionStrategy::WeightedRRF { weights: Vec<f32>, k: f32 }` implementing `Σᵢ weight_i / (rank_i + k)`. Weighted, 0-based variant of RRF for hybrid dense+text search where branches carry different retrieval precision. `text.rs` TODO resolved.

### REST + TypeScript SDK parity (relate / unrelate / relations / TTL)

Four new REST endpoints and TypeScript SDK methods:
- `POST /collections/{name}/relations` — create a relation edge
- `DELETE /collections/{name}/relations/{edge_id}` — remove a relation edge
- `GET /collections/{name}/points/{id}/relations` — list outgoing edges
- `PATCH /collections/{name}/points/{id}/ttl` — set durable TTL

Server handlers in `crates/velesdb-server/src/handlers/points/relations.rs` cover all `AnyCollection` types. TypeScript SDK has HTTP backend (`graph-backend.ts`), WASM stubs, client validation layer, and backend interface additions.

### GraphFirst hybrid anchor + ORDER BY similarity() exhaustiveness

`hybrid_search_with_anchors` exposes a pre-filtered hybrid path for GraphFirst queries with `AND NEAR + MATCH` predicates, ensuring anchor candidates outside the global top-K are not missed. ORDER BY similarity() without a NEAR anchor now falls back to exhaustive payload-resolved scoring.

### Parallel planner (`ExecutionStrategy::Parallel`)

`dispatch_near_with_filter` runs GraphFirst + VectorFirst independently then union-merges by best score per ID, enabling hybrid retrieval when both graph and vector signals are available.

### HnswParams.alpha propagation to native graph

`new_with_storage_mode()` replaced by `new_with_options()` to thread `params.alpha` into the native HNSW graph constructor, so tuned alpha values are respected instead of always using the default.

### PQ codebook persistence on every full flush

`flush_core_storage` now calls `flush_pq_codebook()` after saving the HNSW index, ensuring lazy-trained quantizers survive restart. Also saves immediately after lazy training in `crud_helpers.rs` so the codebook is durable without waiting for the next full flush. The `drain_delta_into_index` / `drain_deferred_into_index` duplication is resolved by extracting `filter_and_insert_valid`.

### TTL TOCTOU fix in MATCH executor

`now_secs` is now snapshotted once before the result loop in `similarity.rs` and used to skip expired points inline. Previously the clock was sampled per-result, creating a TOCTOU window.

### Gap recovery skip when index is complete

`recover_hnsw_gap` emits a `debug!` log and returns early when `storage_count == 0` or `storage_count == hnsw_count`, making no-op recovery observable. New test `test_no_gap_after_flush_and_reopen` asserts zero gap after a clean flush+restart.

---

## Quality checklist

- [x] Zero `unwrap()`/`expect()` in production code
- [x] All new `unsafe`-free
- [x] `cargo clippy --workspace --all-targets --features persistence,gpu,update-check -- -D warnings -D clippy::pedantic` clean
- [x] `cargo fmt --all -- --check` clean
- [x] `cargo test -p velesdb-core --features persistence -- --test-threads=1` — 0 failures
- [x] `npx tsc --noEmit` clean (TypeScript SDK)
- [x] `npx vitest run` — 778 pass / 0 fail (TypeScript SDK)
- [x] `python3 scripts/check_prod_unwraps.py` — PASSED
- [x] `python3 scripts/verify_unsafe_safety_template.py` — PASSED
- [x] NLOC ≤ 50, CCN ≤ 8 on all changed functions

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -256,6 +256,10 @@ engines:
       # in a struct would add indirection without clarity gain.
       - "crates/velesdb-core/src/collection/search/query/execution_paths.rs"
       - "crates/velesdb-core/src/collection/search/query/join.rs"
+      # select_dispatch.rs: `resolve_initial_results` (10 params) threads per-query
+      # context through the anchored/hybrid-anchored/vector-fallback dispatch chain;
+      # same rationale as execution_paths.rs / join.rs above.
+      - "crates/velesdb-core/src/collection/search/query/select_dispatch.rs"
 
       # --- CLI / REPL (interactive tool, not production engine) ---
       - "crates/velesdb-cli/src/**"
--- a/crates/velesdb-core/src/api_types/responses.rs
+++ b/crates/velesdb-core/src/api_types/responses.rs
@@ -176,10 +176,22 @@ pub enum QueryType {
 /// - `SELECT *` → `{id, field1, field2, ...}` (no vector)
 /// - `SELECT col1, col2` → `{col1, col2}`
 /// - `SELECT similarity() AS score, title` → `{score, title}`
+///
+/// ## MATCH-query extra fields
+///
+/// When the query contains a `MATCH` clause, each row may carry additional
+/// graph-specific fields:
+/// - `_edge_bindings` — map of alias → edge (id, source, target, label, properties)
+///   for every named relationship in the pattern.
+/// - `_edge_paths` — array of edge-ID arrays, one per matched path variable.
+///   Only present when the `MATCH` clause binds a path variable (`p = (…)-[…]->(…)`).
 #[derive(Debug, Serialize)]
 #[cfg_attr(feature = "openapi", derive(ToSchema))]
 pub struct QueryResponse {
     /// Projected result rows. Shape depends on SELECT clause.
+    ///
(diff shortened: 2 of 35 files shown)


cyberlife-coder:

PR #1108, opened 2026-06-14
fix(wasm): overflow-safe bounds checks + chunks_exact in multi_query_search_impl / batch_search_impl

cyberlife-coder:
## What

Two correctness bugs in `crates/velesdb-wasm/src/store_search.rs` found during the PR review cycle (2026-06-14):

### Bug 1 — `multi_query_search_impl`: unchecked multiplication in bounds validation (overflow UB)

```rust
// Before — overflows usize on wasm32 when num_vectors × dimension > u32::MAX
if vectors.len() != num_vectors * dimension { … }
```

On 32-bit targets (`wasm32-unknown-unknown`, `usize = u32`) large inputs make `num_vectors * dimension` wrap, producing a wrong sentinel that passes validation silently. The function then slices the buffer with the same wrong start indices.

### Bug 2 — `batch_search_impl`: missing bounds check entirely

`batch_search_impl` had **no** length validation. A caller passing `num_vectors=3` with `vectors=[0.0; 4]` and `dimension=4` would attempt `vectors[4..8]` → panic at runtime. The public `batch_search` entry point in `vector_store.rs` validated before calling `batch_search_impl`, but that validation also used unchecked multiplication.

## Fix

- Extract `validate_multi_vector_len(len, num_vectors, dim) -> Result<usize, String>` — uses `checked_mul` so the product is overflow-safe; returns a `String` error that both WASM functions and native unit tests can consume without touching `JsValue`.
- `multi_query_search_impl` delegates to the helper and replaces the manual `for i in 0..num_vectors { let start = i * dimension; … }` loop with `vectors.chunks_exact(dimension)`, removing a **second** unchecked multiplication in the hot path.
- `batch_search_impl` gains the missing bounds check (via the same helper) and uses `chunks_exact` for both query and store-data iteration.
- `VectorStore::batch_search` (public WASM entry point) delegates its length check to `validate_multi_vector_len` — DRY, no duplicate expression.

## Why chunks_exact instead of manual indexing

`chunks_exact(n)` is:
- **Panic-free** for the iteration itself — no `[start..start+dim]` arithmetic that could independently overflow
- **Idiomatic** — standard Rust pattern for fixed-size chunking
- **Cleaner** — eliminates the loop counter `i` whose only role was computing `start`

## Tests (native, no WASM runtime required)

`validate_multi_vector_len` returns `String` errors, so it can be tested in `cargo test --lib`:

| Test | Asserts |
|---|---|
| `test_validate_multi_vector_len_length_mismatch` | 4 floats declared as 2×dim4 → `Err` with "size mismatch" |
| `test_validate_multi_vector_len_exact_match` | 8 floats = 2×dim4 → `Ok(8)` |
| `test_validate_multi_vector_len_zero_vectors` | 0 vectors × dim4 → `Ok(0)` |
| `test_validate_multi_vector_len_overflow_returns_error` | `usize::MAX` × 2 → `Err` with "overflow" (no panic) |

All 4 new tests + 523 existing WASM lib tests pass on native (`cargo test -p velesdb-wasm --no-default-features --lib`).

## Relationship to open PRs

- PR #1107 touches `store_search.rs` for `validate_dimension` only — **different functions**, no conflict.
- No other open PR addresses `multi_query_search_impl`, `batch_search_impl`, or `VectorStore::batch_search`.

## Merge order

This fix is independent of the 7 open PRs and can be merged at any point. When PR #1107 lands, git will merge `validate_dimension` changes and `validate_multi_vector_len`+`chunks_exact` changes cleanly (they are in different functions of the same file).

https://claude.ai/code/session_01GVw6dcSyeaiskrcqfLAmtK

### DIFF
--- a/crates/velesdb-wasm/src/lib_tests.rs
+++ b/crates/velesdb-wasm/src/lib_tests.rs
@@ -283,3 +283,57 @@ fn test_similarity_threshold_logic() {
     assert!((0.8001_f32 - threshold).abs() < 0.001);
     assert!((0.81_f32 - threshold).abs() >= 0.001);
 }
+
+// =============================================================================
+// Bounds-check and overflow safety — validate_multi_vector_len
+//
+// The JsValue-returning search functions are wasm-only and cannot be called
+// from native tests (JsValue::from_str aborts on non-wasm32). The validation
+// logic is extracted into validate_multi_vector_len (returns String errors) so
+// it can be exercised here. The full wasm integration path is covered by the
+// wasm-pack test suite in CI.
+// =============================================================================
+
+#[test]
+fn test_validate_multi_vector_len_length_mismatch() {
+    use super::store_search::validate_multi_vector_len;
+    // 4 floats declared as 2 vectors of dim 4 — expected 8, got 4.
+    let err = validate_multi_vector_len(4, 2, 4);
+    assert!(err.is_err(), "length mismatch must be an error");
+    let msg = err.unwrap_err();
+    assert!(
+        msg.contains("size mismatch"),
+        "error should describe the mismatch: {msg}"
+    );
+}
+
+#[test]
+fn test_validate_multi_vector_len_exact_match() {
+    use super::store_search::validate_multi_vector_len;
+    // 8 floats = 2 vectors × dim 4: valid.
+    let ok = validate_multi_vector_len(8, 2, 4);
+    assert!(ok.is_ok(), "matching lengths must succeed");
+    assert_eq!(ok.unwrap(), 8);
+}
+
+#[test]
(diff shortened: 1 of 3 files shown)


cyberlife-coder:
Thanks for catching the unchecked `num_vectors * dimension` — the overflow concern is valid. I can't merge this PR as-is: the commit authorship doesn't satisfy the repository CLA (hence the failing `cla-assistant` check) and the project requires contributions under the maintainer's authorship.

I've reimplemented the same fix (`checked_mul` + `chunks_exact`, shared helper, unit tests) in #1109, which supersedes this. Closing in its favor.

PR #1137, opened 2026-06-15
refactor(migrate+sdk/ts): remove dead allow(deprecated), modernise guards, harden TypeScript types

cyberlife-coder:
$(cat <<'EOF'
## Summary

Targeted quality sweep based on a full codebase audit (2026-06-15). All changes are pure refactors — no behaviour change, no new dependencies.

### Rust (`velesdb-migrate`)

**`pipeline_points.rs`**
- Remove vestigial `#[allow(deprecated)] // TODO(MIGRATE-01)` on the `get_vector_collection()` call. The migration to the non-deprecated method was already done; only the suppressor remained, producing confusing dead code.

**`pipeline_graph.rs`**
- `attach_weight`: replace dual `match … return` guards with `let…else` (Rust 1.65, stabilised Nov 2022). Reduces two match blocks to two one-line guards, making early-return intent explicit at the binding site.
- `build_edge`: rewrite error path with `Result::inspect_err(…).ok()?` (Rust 1.76). Logs the edge-creation failure and converts to `Option` in a single chain, removing a four-line match block.
- `attach_weight`: replace `HashMap::new()` + `insert` with `HashMap::from([…])` for the single-entry weight map. Eliminates a `mut` binding.

### TypeScript SDK (`sdks/typescript`)

**`shared.ts`**
- Extract `safeJsonParse(response)` helper — consolidates the `.response.json().catch(() => ({}))` pattern used at 5 HTTP error-body sites. Return type is `Record<string, unknown>`, which is more accurate than `any` and compatible with all `typeof data.code === 'string'` type-narrowing patterns callers already use.

**`crud-backend.ts` · `rest-http.ts` · `streaming-backend.ts`**
- Replace all 5 occurrences of the raw `.catch(() => ({}))` with `safeJsonParse(response)`.

**`query-backend.ts`**
- Drop `#@typescript-eslint/no-explicit-any` + `as any` cast on `response.data`. The generic `requestJson<Record<string, unknown>>` already types the field correctly; explicit `as number | undefined` casts on the scalar properties replace the implicit `any` widening.

**`search-backend.ts` · `streaming-backend.ts`**
- Change 4 request-body objects from `Record<string, any>` to `Record<string, unknown>`. These are write-only construction patterns; `unknown` is the correct type for values being sent as JSON. Removes 4 `eslint-disable` suppression comments.

## Test plan

- [ ] Rust: `cargo test -p velesdb-migrate` — all existing pipeline tests pass
- [ ] TypeScript: `cd sdks/typescript && npm test` — all existing vitest tests pass  
- [ ] TypeScript: `cd sdks/typescript && npm run typecheck` — zero type errors with the stricter types
- [ ] CI: all existing CI gates (Lint & Format, Tests, TypeScript SDK Check) pass

> Factual as of 2026-06-15. Rust MSRV 1.89 (workspace). `let…else` requires ≥ 1.65; `inspect_err` requires ≥ 1.76; both well within MSRV.

https://claude.ai/code/session_01LoXMNLw5CobZGiefETWnMw
EOF
)

### DIFF
--- a/crates/velesdb-migrate/src/pipeline_graph.rs
+++ b/crates/velesdb-migrate/src/pipeline_graph.rs
@@ -177,14 +177,9 @@ fn build_edge(
     let key = format!("{from_node_id}-{to_node_id}-{}", relation.edge_label);
     let id = crate::pipeline::fnv1a64(key.as_bytes());
 
-    let edge =
-        match velesdb_core::GraphEdge::new(id, from_node_id, to_node_id, &relation.edge_label) {
-            Ok(e) => e,
-            Err(e) => {
-                warn!("Failed to create edge {}: {}", id, e);
-                return None;
-            }
-        };
+    let edge = velesdb_core::GraphEdge::new(id, from_node_id, to_node_id, &relation.edge_label)
+        .inspect_err(|e| warn!("Failed to create edge {}: {}", id, e))
+        .ok()?;
 
     Some(attach_weight(edge, point, relation))
 }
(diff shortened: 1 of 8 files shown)


cyberlife-coder:
Superseded by #1139, which carries the same diff on a Git-Flow-compliant `refactor/*` branch with maintainer-authored commits (the commits here were authored under a non-signatory identity, failing the CLA check, and the source branch prefix is not allowed to target develop).

PR #1155, opened 2026-06-17
perf/refactor(migrate): reduce allocations and eliminate code duplication in connectors

cyberlife-coder:
## Summary

Three atomic improvements to the `velesdb-migrate` crate, all following Rust 2026 best practices (zero unnecessary allocations, DRY, no behaviour changes).

### Commit 1 — `perf(migrate): avoid Vec<u8> clone in RESP string extraction`

`extract_bulk_string` and `resp_value_to_json` in `redis.rs` called `String::from_utf8(bytes.clone())`, which allocates a full copy of the byte slice before discovering whether it is valid UTF-8. On the error path that allocation was immediately discarded.

**Fix:** use `std::str::from_utf8(bytes)` — zero allocation on the rejection path; only `str::to_owned()` allocates when the bytes are valid UTF-8, which is the case that produces the returned `String` anyway.

### Commit 2 — `refactor(migrate): DRY payload extraction via extract_payload_from_hashmap`

`build_payload` in `redis.rs` and `extract_payload_from_object` in `common.rs` both filter and clone key-value pairs into an owned `HashMap` — duplicated logic with a subtle cost in the test helper path:

- `extract_payload` (test helper) wrapped the `HashMap` in a `serde_json::Value::Object`, cloning every entry **twice** — once to build the Object, once inside `extract_payload_from_object`.

**Fix:** add `extract_payload_from_hashmap` to `common.rs` that accepts any `IntoIterator<Item = (&String, &Value)>` (directly covering the HashMap case). `extract_payload_from_object` becomes a thin wrapper. `build_payload` and the test helper both delegate: one clone pass instead of two in the test path.

Clippy `manual_contains` lint that arose from the new iterator (use `slice.contains(&x)` rather than `.iter().any(|f| f == x)`) fixed in the same commit.

### Commit 3 — `refactor(migrate): eliminate duplicate index-URL logic and Option clone`

Two cleanups found during the connector sweep:

1. **Elasticsearch** — `fetch_field_similarity` manually reproduced the exact same 7-line URL-building logic already in `build_index_url` (trim trailing slash, handle leading-slash index names, concat action suffix). Replace the duplicate block with `self.build_index_url("_mapping")`.

2. **Pinecone** — `control_plane_url` called `self.config.base_url.clone().unwrap_or_else(…)`, cloning the entire `Option<String>` before unwrapping it. Use `as_deref().unwrap_or(…).to_string()` — the `Some` path borrows the existing heap string, the `None` path returns a static default: one allocation either way instead of two.

## Test plan

- [x] `cargo test -p velesdb-migrate --features redis-source` — 15/15 pass (includes Elasticsearch, Pinecone, Qdrant, Supabase, JSON e2e, dry-run, checkpoint resume, sparse vectors, pagination, error paths)
- [x] `cargo clippy -p velesdb-migrate --all-targets -- -D warnings` — clean
- [x] `cargo check --workspace --exclude tauri-plugin-velesdb` — clean
- [x] No behaviour changes — all public APIs unchanged, all test assertions unchanged

https://claude.ai/code/session_01VHvYgJZxpb77o5qtHL7y3P

### DIFF
--- a/crates/velesdb-migrate/src/connectors/common.rs
+++ b/crates/velesdb-migrate/src/connectors/common.rs
@@ -248,23 +248,28 @@ pub fn extract_payload_from_object(
     excluded_fields: &[&str],
     included_fields: &[String],
 ) -> HashMap<String, Value> {
-    let mut payload = HashMap::new();
+    let Value::Object(map) = source else {
+        return HashMap::new();
+    };
+    extract_payload_from_hashmap(map, excluded_fields, included_fields)
+}
 
-    if let Value::Object(map) = source {
-        for (key, val) in map {
-            // Skip excluded fields
-            if excluded_fields.iter().any(|f| f == key) {
-                continue;
-            }
-            // If included_fields is specified, only include those
-            if !included_fields.is_empty() && !included_fields.contains(key) {
-                continue;
-            }
-            payload.insert(key.clone(), val.clone());
-        }
-    }
-
-    payload
+/// Extracts payload fields from a pre-parsed `HashMap<String, Value>`.
+///
+/// Identical filtering semantics to [`extract_payload_from_object`] but
+/// operates directly on a `HashMap`, avoiding a round-trip through
+/// `serde_json::Value::Object` when the caller already has the map form.
+pub fn extract_payload_from_hashmap<'a>(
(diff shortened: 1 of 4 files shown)


cyberlife-coder:

PR #1197, opened 2026-06-22
fix(security): add input-bounds limits [superseded — see fix/input-bounds-limits]

cyberlife-coder:
## Summary

Adds missing point-count and sparse-NNZ guards, consistent with the existing `MAX_SCROLL_BATCH_SIZE` (10 K) and `MAX_BULK_DELETE_SIZE` (10 K) limits, on the two ingest paths that had none.

### `MAX_UPSERT_BATCH_SIZE = 100 000`

- Enforced in **JSON upsert** (`POST /collections/{name}/points`) — checked before `build_points_from_request`, returns `400 BAD_REQUEST` with a clear message.
- Enforced in **binary VRB1 raw upsert** (`POST /collections/{name}/points/raw`) — checked after `vrb1::decode`, same response.
- Ingestion workloads legitimately need larger batches than read/delete (hence 100 K vs 10 K), but without any limit a metadata-only or tiny-vector corpus can fit millions of rows in a single 100 MB body, causing unbounded memory growth during deserialization and HNSW insertion.

### `MAX_SPARSE_NNZ = 65 536` — in `velesdb-core::api_types::requests`

- Checked in `SparseVectorInput::convert_parallel` and `convert_dict` before the `Vec<(u32, f32)>` allocation.
- SPLADE / BM25 embeddings typically have < 1 K NNZ; 64 K gives 60× real-world headroom while bounding worst-case heap use to ~512 KB per sparse vector. Without this a single parallel-format sparse vector with every entry set costs 100 MB / (2 bytes per JSON char × 2 entries per pair) ≈ 25 M NNZ → ~200 MB heap.

### OpenAPI spec sync

Also regenerates `docs/openapi.json` / `docs/openapi.yaml` which had pre-existing drift: edge/node ID schemas and the `400` description on the JSON upsert endpoint.

## Test plan

- [x] `cargo check --package velesdb-server --package velesdb-core` — clean
- [x] `cargo test --package velesdb-server --lib` — 148 passed, 0 failed
- [x] `cargo test --package velesdb-core --lib api_types` — 81 passed, 0 failed
- [x] New tests: `handlers::points::tests` (constant-value sanity + ordering invariant)
- [x] New tests: `api_types::requests::tests` (oversized-NNZ rejection × 2, max-NNZ acceptance, length-mismatch, non-finite rejection)

## Security impact (2026-06-22)

| Path | Before | After |
|------|--------|-------|
| `POST /points` JSON | ∞ points | 100 000 |
| `POST /points/raw` VRB1 | ∞ points | 100 000 |
| Sparse NNZ | ∞ | 65 536 |
| `POST /points/scroll` | 10 000 (existing) | unchanged |
| `POST /points/delete` | 10 000 (existing) | unchanged |

https://claude.ai/code/session_01ESgTviGUEHaB8DkAPZi6ZQ

### DIFF
--- a/crates/velesdb-core/src/api_types/requests.rs
+++ b/crates/velesdb-core/src/api_types/requests.rs
@@ -142,13 +142,21 @@ pub enum SparseVectorInput {
     Dict(BTreeMap<String, f32>),
 }
 
+/// Maximum non-zero entries (NNZ) allowed in a single sparse vector.
+///
+/// Prevents memory amplification: a 100 MB JSON body of sparse indices/values
+/// can decompress to a much larger in-memory `Vec<(u32, f32)>`. Sparse NLP
+/// embeddings (SPLADE, BM25) typically have < 1 K NNZ; 65 536 gives a 60× headroom
+/// over typical workloads while bounding worst-case allocation to ~512 KB per vector.
+pub const MAX_SPARSE_NNZ: usize = 65_536;
+
 impl SparseVectorInput {
     /// Converts this input into a core `SparseVector`, validating at the API boundary.
     ///
     /// # Errors
     ///
     /// Returns a descriptive error string on mismatched lengths, non-finite values,
-    /// or dict keys that cannot be parsed as `u32`.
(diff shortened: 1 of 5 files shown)


cyberlife-coder:

PR #1280, opened 2026-06-29
fix(memory): server.json description <=100 chars (MCP registry validation)

cyberlife-coder:
The MCP registry rejects `description` > 100 chars (HTTP 422 on publish). Shortens server.json description to 99 chars while still leading with **agentic memory** + the **tri-engine** (vector+graph+columnar). Unblocks the publish-mcp-registry workflow added in #1279.

Tested: the OIDC auth + namespace resolution already succeed (the failed run got to validation); only the length was rejected.

### DIFF
--- a/server.json
+++ b/server.json
@@ -2,7 +2,7 @@
   "$schema": "https://static.modelcontextprotocol.io/schemas/2025-12-11/server.schema.json",
   "name": "io.github.cyberlife-coder/velesdb-memory",
   "title": "VelesDB Memory",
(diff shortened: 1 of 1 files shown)


cyberlife-coder:
Superseded by #1281 (mcpb pipeline): the MCP registry doesn't support registryType=cargo, so server.json is now generated dynamically in CI as mcpb packages. The ≤100-char description constraint is carried into the generated server.json. Closing — the static cargo server.json is legacy/Glama-facing only.

PR #1301, opened 2026-07-01
feat(memory): add --dump JSONL instrumentation to the LoCoMo harness

cyberlife-coder:
## Summary
- Adds `--dump <path>` to `examples/locomo/` (velesdb-memory): appends one JSONL `QuestionRecord` per QA per mode (vector-only / fused), needed for the temporal-decomposition research analysis (why the scaffold costs -6pp single-hop) — no per-question data exists today, only aggregate percentages.
- Each record carries: predicted/gold/correct/f1, date/scaffold/prompt-kind flags, `is_temporal_trigger`, token usage (from Ollama's own reply when available, honestly `None` on a cache hit or Claude-gen), a whitespace-estimate `context_tokens` fallback, date-span stats, and both the pre-budget candidate pool (`raw_facts`) and the final budgeted set (`reranked_facts`) so "never retrieved" can be told apart from "retrieved but evicted by ranking".
- Purely observational: built from data `eval::evaluate` already computes; no second retrieval call, no change to any prompt string. Ollama's cache key is `FNV1a(model + prompt)` (`ollama_gen.rs`), so it is unaffected by `--dump`.
- Scope: **this PR is instrumentation only** — no dataset run, no JSONL committed, no analysis. The actual 3-config 10-conversation reproduction run (expensive: hours of local generation + Claude-judge API calls) and the statistical/taxonomy analysis are a separate follow-up, gated on review of this PR.

## Verification
- [x] `cargo test -p velesdb-memory --features ollama --example locomo`: 6/6 new tests green (JSONL round-trip, date-span math, evidence-hit split, token estimate).
- [x] `cargo test -p velesdb-memory` (full crate, default features): all green — no lib behavior touched.
- [x] `cargo +1.89 fmt --check` / `cargo clippy --profile test -p velesdb-memory --features ollama --example locomo -- -D warnings -D clippy::pedantic`: clean.
- [x] `lizard` (CCN≤8, NLOC≤50): no new violations (the one pre-existing violation, `judge::answer`, predates this change and its shape is untouched here).
- [x] Live smoke run (1 conversation, 2 QA, local `qwen3.6:35b-mlx` + `all-minilm`): confirmed byte-identical cache-file count and identical predicted answers/report table with `--dump` on vs off (cold + warm reruns); confirmed `prompt_tokens`/`completion_tokens` populate on a cold call and are honestly `None` on a cache hit; confirmed `reranked_facts` count == budget `k` and `raw_facts` == the oversampled pre-budget pool.
- [x] Pre-commit hook passed in full (fmt, clippy, full workspace `--lib` test suite incl. velesdb-core, SAFETY/TODO governance, secret scan).

## Follow-up (not in this PR)
The 3-config faithful-reproduction run, statistical analysis (McNemar/Wilson/bootstrap), error taxonomy, and the Smart Routing chapter are the next phases of the durable research plan — deliberately deferred pending review of this instrumentation.

### DIFF
--- a/crates/velesdb-memory/examples/locomo/dump.rs
+++ b/crates/velesdb-memory/examples/locomo/dump.rs
@@ -1,0 +1,239 @@
+//! Per-question JSONL instrumentation for the `LoCoMo` temporal-decomposition
+//! research analysis (`--dump <path>`).
+//!
+//! Purely observational: every field is read from data `eval.rs` already
+//! computes for the real vector/graph/generation path. There is no second
+//! retrieval call and no change to any prompt string, so the generation/judge
+//! cache (content-addressed on `model + prompt`, see `ollama_gen.rs`) is
+//! identical whether `--dump` is on or off.
+
+use std::error::Error;
+use std::fs::File;
+use std::io::{BufWriter, Write as _};
+use std::path::Path;
+
+use serde::Serialize;
+
+use crate::dataset::Qa;
+use crate::eval::{EvalCfg, RetrievedFact};
+use crate::ollama_gen::TokenUsage;
+
+/// One retrieved fact, as it appeared in either the pre-budget candidate pool
+/// (`raw_facts`) or the final budgeted set the generator actually saw
+/// (`reranked_facts`). Comparing the two on `evidence_hit` tells "never
+/// retrieved" (missing from both) apart from "retrieved but evicted by
+/// ranking" (present in `raw_facts` only).
+#[derive(Serialize)]
+pub struct FactRecord {
+    pub rank: usize,
+    pub id: u64,
+    pub ts: i64,
+    pub vector_score: f64,
+    pub graph_weight: f64,
+    pub evidence_hit: bool,
+    pub dia_ids: Vec<String>,
+    pub text: String,
+}
+
+/// One question's full evaluation trace under one mode (vector-only or fused).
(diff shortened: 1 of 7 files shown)


cyberlife-coder:
Ran an adversarial review (8 independent angles + 1-vote verification) before merging. Found and fixed 2 correctness bugs in the dump instrumentation itself, in this follow-up commit:

- `raw_if_wanted` wasn't deduplicating a fact that's both vector-ranked and graph-reached (unlike `fuse()`, which already does), so it appeared twice in `raw_facts` with split scores.
- `fuse()` never synced the resolved graph weight back onto the surviving `RetrievedFact` struct for a fact present in both the vector pool and the graph-reached set — it kept the pool's hardcoded `graph_weight: 0.0` even though its rank used the real weight. Pre-existing in `fuse()`, but inert until this PR started serializing `graph_weight`; a graph-promoted fact looked indistinguishable from a pure vector hit in the dump.

Both are observation-only fixes (scoring/sort order/QA accuracy unaffected — verified `fused_score()` reads from a separate weights map, not the struct field). Also factored the duplicated YYYYMMDD decomposition (`dump.rs`'s `ymd_to_ordinal` vs `judge.rs`'s `fmt_date`) into one shared `decompose_ymd`. 4 new regression tests added. Full test suite + clippy pedantic + fmt re-verified clean.

Merging now.

PR #1434, opened 2026-07-17
feat(python): compile_context + context tools parity on MemoryService [EPIC-P-071/US-001]

cyberlife-coder:
## Summary

- Python parity for the context compiler (EPIC-P-070/US-002's memory bridge): `MemoryService` gains `compile_context`, `retrieve_context_source`, `context_savings`, `save_working_context`, `load_working_context` — the same 5 operations the MCP server and the Node binding already expose.
- Zero new logic: every method is a thin dict↔serde conversion around `velesdb_memory::context::MemoryService::*` (the exact bridge the MCP tools call), using the file's existing manual `json_to_python`/`python_to_json` helpers.
- Fixes those two shared helpers to round-trip a `u64` outside `i64`'s range instead of silently falling through to a lossy `f64` — required so `compile_context`'s ids (FNV-1a 64-bit `stable_id` hashes, uniform over `u64`, so ~50% exceed `i64::MAX`) cross as exact native Python ints. Confirmed against a real value in manual testing: `ctx://source/18021940868160883968`.
- Enables the `velesdb-memory` `context` feature for `velesdb-python` (`context = []`, zero new dependencies — same feature set `velesdb-node` already builds with); without it `compile_context` and friends are unreachable at all.
- README: new "Context Compiler" section with a real executed round trip, plus LangChain and LlamaIndex integration snippets — both actually run against the built extension in this session, output pasted verbatim (no unexecuted code documented).
- `__init__.pyi`: stub signatures + docstrings for the 5 new methods (not otherwise enforced by `test_stub_parity.py`, but keeps typed callers accurate).

## Documented design decisions (per plan)

- Ids cross as **native Python ints** (unlimited precision), not decimal strings like the Node binding — a deliberate, documented wire-shape difference (semantic parity, not literal parity).
- Errors map through the file's existing `to_py_err`/`ErrorCategory` taxonomy — no new exception types. `token_budget=0` → `MemoryError::ContextBudget` → `ValueError`; unknown/malformed `ctx://source/` handle → `MemoryError::UnknownHandle` → `KeyError`.
- `MemoryService` is re-exported as-is by `python/velesdb/__init__.py` (no wrapper class), so the 5 new `#[pymethods]` are automatically visible there — verified, no adapter change needed.

## TDD evidence

RED (`tests/test_context_compiler.py`, before implementation — all 13 fail with `AttributeError`, methods don't exist yet):

```
FAILED tests/test_context_compiler.py::test_compile_context_minimal_round_trip
FAILED tests/test_context_compiler.py::test_compile_context_verbatim_and_cache_fragments
FAILED tests/test_context_compiler.py::test_retrieve_context_source_round_trips_a_compiled_fragment
FAILED tests/test_context_compiler.py::test_retrieve_context_source_unknown_handle_raises_key_error
FAILED tests/test_context_compiler.py::test_retrieve_context_source_malformed_handle_raises_key_error
FAILED tests/test_context_compiler.py::test_context_savings_aggregates_after_a_compile
FAILED tests/test_context_compiler.py::test_context_savings_with_no_project_filter_returns_a_dict
FAILED tests/test_context_compiler.py::test_save_and_load_working_context_round_trips
FAILED tests/test_context_compiler.py::test_save_working_context_is_an_idempotent_upsert
FAILED tests/test_context_compiler.py::test_load_working_context_returns_none_when_absent
FAILED tests/test_context_compiler.py::test_compile_context_zero_budget_raises_value_error
FAILED tests/test_context_compiler.py::test_compile_context_result_shape_matches_the_mcp_wire_contract
FAILED tests/test_context_compiler.py::test_compile_context_ids_survive_full_u64_precision
13 failed in 0.41s
```

GREEN (after implementation):

```
13 passed in 0.64s
```

Full suite (non-regression): `505 passed, 23 skipped` (skip count unchanged from baseline).

## Gates (all green)

```
cargo fmt --all --check                                    # OK
cargo clippy -p velesdb-python -- -D warnings               # OK, zero warnings
RUSTFLAGS=-Dwarnings cargo check -p velesdb-python           # OK
  (adapted from the plan's `--features persistence,ollama,extract`:
   those are velesdb-memory's features, not velesdb-python's own —
   velesdb-python's own feature list is default/gpu/update-check/loom;
   it pulls velesdb-memory with persistence+ollama+extract+context
   unconditionally via its Cargo.toml, so plain `cargo check -p velesdb-python`
   already exercises that full feature set)
cd crates/velesdb-python && pytest tests/ -q                # 505 passed, 23 skipped
```

`cargo test -p velesdb-python` was not run as a gate (not in the plan's gate list) — it fails to link on this machine even on a clean baseline checkout (pre-existing: PyO3 `extension-module` cdylib crates can't resolve `libpython` symbols outside an actual Python process on macOS; confirmed by stashing this PR's changes and re-running the same command against `develop`).

## Test plan

- [x] RED captured before implementation (13/13 fail with `AttributeError`)
- [x] GREEN after implementation (13/13 pass)
- [x] Full `pytest tests/` non-regression (505 passed, 23 skipped, same skip count as baseline)
- [x] `cargo fmt`, `clippy -D warnings`, `cargo check` all clean
- [x] README LangChain/LlamaIndex snippets actually executed against the built extension
- [x] `docs/openapi.json`/`.yaml` untouched (never regenerated by this build)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Node** (`@wiscale/velesdb-memory-node`): `compileContext(request)` —
   pure conversion, ids as decimal strings; ships the
   `velesdb-context-optimizer` agent skill in the npm package.
+- **Python** (`velesdb`): context-compiler parity on `MemoryService` —
+  `compile_context` / `retrieve_context_source` / `context_savings` /
+  `save_working_context` / `load_working_context`, same wire shapes as the
+  MCP tools; ids cross as exact native Python ints (vs decimal strings on
+  Node). [EPIC-P-071/US-001]
 - **Benchmark**: `examples/context_savings`, reproducible (75–82 % estimated
   token savings on the committed corpus in ~2 ms; figures are local
   estimates, not billed tokens — cross-checked against a real cl100k
@@ -75,6 +80,17 @@ account).
 - **Graph cluster held as a shared `Arc<GraphStore>` handle** (R1.2b —
   internal-only): graph state can be shared with `GraphCollection` without an
   exclusive move; field access and locking are unchanged.
+
+### Fixed
+
+- **python: integers greater than `i64::MAX` now round-trip exactly as `int`**
+  (previously silently converted to a lossy `float`). Affects every surface
+  sharing the binding's JSON converters — payloads, search results, VelesQL
+  params, graph properties, memory metadata — and is required by the context
+  compiler's content-addressed ids (FNV-1a 64, uniform over `u64`, so ~50%
+  exceed `i64::MAX`). Integers outside `[i64::MIN, u64::MAX]` now raise an
+  explicit `ValueError` stating the supported range instead of silently losing
+  precision. [EPIC-P-071/US-001]
 
 ## [3.12.0] — 2026-07-15
 
--- a/crates/velesdb-python/Cargo.toml
+++ b/crates/velesdb-python/Cargo.toml
@@ -21,8 +21,11 @@ crate-type = ["cdylib"]
 velesdb-core = { workspace = true }
 # The high-level memory wedge (MemoryService: remember/recall/relate/forget/why
 # + remember_extracted). `default-features = false` drops the MCP server stack
-# (rmcp/tokio-server); `ollama`+`extract` enable real embeddings and extraction.
-velesdb-memory = { path = "../velesdb-memory", default-features = false, features = ["persistence", "ollama", "extract"] }
+# (rmcp/tokio-server); `ollama`+`extract` enable real embeddings and extraction;
+# `context` enables the deterministic context compiler (EPIC-P-071/US-001) —
+# zero new dependencies (`context = []` in velesdb-memory), same feature set
+# velesdb-node already builds with.
+velesdb-memory = { path = "../velesdb-memory", default-features = false, features = ["persistence", "ollama", "extract", "context"] }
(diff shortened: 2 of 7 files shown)


cyberlife-coder:
Review follow-ups applied (3 commits pushed):

1. **CHANGELOG (reserve)** — `[Unreleased]` now carries a `### Fixed` entry for the binding-wide u64 behavior change ("python: integers greater than i64::MAX now round-trip exactly as int; previously silently converted to lossy float"), plus a Python bullet in the compiler Added block alongside the existing Node one.

2. **utils.rs upper bound** — any Python int outside `[i64::MIN, u64::MAX]` now raises an explicit `ValueError` naming the supported range, before the `f64` fallback can silently round it. TDD: RED first —

   ```
   tests/test_context_compiler.py::test_metadata_int_above_u64_max_raises_value_error FAILED
   tests/test_context_compiler.py::test_metadata_int_below_i64_min_raises_value_error FAILED
   E       Failed: DID NOT RAISE ValueError
   ```

   GREEN after the guard: `19 passed` in the file.

3. **Boundary tests** — parametrized round-trip in `tests/test_context_compiler.py`: `u64::MAX`, `i64::MAX + 1` (exact int), `1.5` (stays float), `True` (stays bool — type-checked, since `True == 1`), plus `2**64` and `-(2**63)-1` → `ValueError`.

Gates re-run after the fixes: `cargo fmt --all --check` OK, `cargo clippy -p velesdb-python -- -D warnings` OK, `maturin develop --release` + `python -m pytest tests/ -q` → **511 passed, 23 skipped** (505 + 6 new). No openapi files in the diff. `origin/develop` fetched and merged — already up to date (branch base is still develop's tip; #1433 not merged yet).

PR #1438, opened 2026-07-18
feat(memory): wire-safe ids opt-in, positional explain, installable skills bundle [EPIC-P-071]

cyberlife-coder:
## Summary

Wave 5 (5.1 + 5.2 + 5.3, S each) of the EPIC-P-071 backlog solde:

- **5.1 — wire-safe MCP ids (opt-in strings).** `CompilePolicy.ids_as_strings` (default `false`) rewrites every id field (`fragment_id`, `content_hash`, `memory_id`, `fragment_ids`) of the `compile_context` / `explain_compilation` MCP response into a decimal string — reuses the existing `context::wire::stringify_id_fields` tree walk (the exact one Node/WASM already apply), no duplication. Protects a raw MCP JS client (no binding) from silent u64 precision loss above `2^53`. Input side: `fragments[].id` now also accepts a decimal string (in addition to a JSON number), via a shared deserializer reusing `wire::parse_u64`.
- **5.2 — `explain_compilation` positional lookup.** New optional `fragment_index` (0-based position in `request.fragments`), in addition to `fragment_id`. Byte-identical fragments share a content-addressed `fragment_id`, so a plain id lookup always resolves to the deduplication survivor's decision — `fragment_index` disambiguates by position instead. Out-of-bounds index → `INVALID_PARAMS` with a clear reason. Default behavior (no `fragment_index`) is unchanged.
- **5.3 — installable skills bundle.** `.github/workflows/release.yml` gains a `package-skills` job that tars `velesdb-memory` + `velesdb-context-optimizer` (one folder per skill at the archive root) into `velesdb-skills.tar.gz` and attaches it to every GitHub Release, so both skills install with `curl -L .../velesdb-skills.tar.gz | tar -xz -C ~/.claude/skills/` — no repo clone required.

## TDD evidence (red → green)

**5.1** — `test_compile_context_tool_ids_as_strings_stringifies_response_ids`, `test_explain_compilation_tool_ids_as_strings_stringifies_response_ids` (RED before the `ids_as_strings` wiring: stringify was stubbed to a no-op, both failed asserting `fragment_id` was still a JSON number). `test_compile_context_tool_ids_as_strings_default_false_is_byte_identical` pins the inactive case. `test_compile_context_tool_accepts_fragment_id_as_decimal_string_on_input` covers the input side.

**5.2** — `test_explain_compilation_tool_fragment_index_disambiguates_byte_identical_twins` (RED before wiring: `fragment_index` lookup was stubbed to always return `None`, so the call errored instead of returning the dropped twin's decision). `test_explain_compilation_tool_fragment_index_out_of_bounds_is_invalid_params` covers the bounds check.

Both reds captured by temporarily stubbing the new code paths, confirmed failing, then restored — full sequence re-run green (17/17 `mcp::context_tools` tests).

**5.3** — verified locally: staged both skill directories into a temp dir, ran the exact `tar -czf ... -C stage velesdb-memory velesdb-context-optimizer` command from the workflow, extracted into a fresh temp dir, confirmed the structure:
```
velesdb-memory/SKILL.md
velesdb-context-optimizer/SKILL.md
```
YAML validated with `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/release.yml'))"`. No release triggered (workflow untouched otherwise; `package-skills` only runs on the existing push-tag / `publish_binaries` dispatch gate, same as the binaries job).

## Gates (all green)

- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets --features persistence,gpu,update-check --exclude velesdb-python --exclude velesdb-node -- -D warnings -D clippy::pedantic`
- `cargo clippy -p velesdb-node --lib -- -D warnings`
- `RUSTFLAGS=-Dwarnings cargo check -p velesdb-memory --no-default-features --features context`
- `cargo test -p velesdb-memory` — 179 unit + 173 BDD/property/doc tests, 0 failed
- `cargo build --release -p velesdb-memory` + `python3 crates/velesdb-memory/examples/context_savings/real_measures/mcp_e2e.py` — OK (6 tools exercised over real stdio)

## Notes / deviations

- `compile_context` / `explain_compilation` now return the wire `Value` directly (instead of `Json<CompiledContext>` / `Json<ContextDecision>`) so the id rewrite can run before serialization completes; the advertised MCP `output_schema` is preserved unchanged via an explicit `output_schema = schema_for_output::<T>()` attribute (byte-identical to what the macro auto-derived before). A handful of existing unit tests were updated to deserialize the returned `Value` back into the typed struct before asserting on fields — same assertions, no behavior change.
- Skill name check: no top-level `skills/velesdb-memory*` exists in this repo — the memory skill lives at `crates/velesdb-memory/skill/velesdb-memory/SKILL.md` (singular `skill/`). The bundle step reads it from that path; `skills/velesdb-context-optimizer/` is unchanged. No `SKILL.md` content was modified, so the two existing copies (`skills/velesdb-context-optimizer` and `crates/velesdb-node/skills/velesdb-context-optimizer`) stay byte-identical, unaffected by this PR.

## Test plan

- [x] `cargo test -p velesdb-memory` green
- [x] `mcp_e2e.py` green against the release-built binary
- [x] tar structure verified locally (one folder per skill at archive root)
- [x] YAML validated
- [ ] CI on this PR (not triggered by this change; branch protection / CI config unaffected)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -239,6 +239,39 @@ jobs:
             *.deb
 
   # ===========================================================================
+  # 2b. Package the agent skills (installable without cloning the repo)
+  # ===========================================================================
+  package-skills:
+    name: Package skills bundle
+    runs-on: ubuntu-latest
+    needs: validate
+    if: github.event_name == 'push' || inputs.publish_binaries
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      - name: Build velesdb-skills.tar.gz
+        # One folder per skill at the archive root, so
+        # `tar -xz -C ~/.claude/skills/` drops each skill straight into
+        # place — the one-line install README documents. Staged in a temp
+        # dir first (the two skills live under different source trees) so
+        # the tar entries carry no leading path noise.
+        run: |
+          set -euo pipefail
+          stage="$(mktemp -d)"
+          cp -r crates/velesdb-memory/skill/velesdb-memory "$stage/"
+          cp -r skills/velesdb-context-optimizer "$stage/"
+          tar -czf velesdb-skills.tar.gz -C "$stage" velesdb-memory velesdb-context-optimizer
+
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          # `binaries-*` on purpose: the GitHub Release job downloads every
+          # artifact matching that pattern and attaches it as-is, so this
+          # rides the same collection step the platform binaries use
+          # without needing a second download/attach path.
+          name: binaries-skills
+          path: velesdb-skills.tar.gz
+
+  # ===========================================================================
   # 3. Publish to crates.io (each crate independently)
   # ===========================================================================
   publish-crates:
@@ -629,12 +662,13 @@ jobs:
   github-release:
     name: GitHub Release
     runs-on: ubuntu-latest
-    needs: [validate, build]
+    needs: [validate, build, package-skills]
     if: |
       always() &&
       needs.validate.result == 'success' &&
       (github.event_name == 'push' || inputs.publish_github) &&
-      (needs.build.result == 'success' || needs.build.result == 'skipped')
+      (needs.build.result == 'success' || needs.build.result == 'skipped') &&
+      (needs.package-skills.result == 'success' || needs.package-skills.result == 'skipped')
     permissions:
       contents: write
     steps:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,6 +121,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (`RUN_BILLED_MEASURE=1` + an API key never asked for by the harness), the
   provider's own billed `usage` on a real minimal-cost call.
   [EPIC-P-071/US-007]
+- **MCP**: `CompilePolicy.ids_as_strings` (default `false`) — opts the
+  `compile_context` / `explain_compilation` response into decimal-string ids
+  (`fragment_id`, `content_hash`, `memory_id`, `fragment_ids`), reusing the
+  same tree walk the Node/WASM bindings already apply, for raw MCP clients
+  without u64-safe JSON number parsing. `fragments[].id` on input now also
+  accepts either a JSON number or a decimal string, and the advertised tool
+  schemas type every such field `["integer", "string"]` so schema-validating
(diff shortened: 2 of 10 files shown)


cyberlife-coder:
Review fixes pushed (c570b4fd) — all three points addressed:

**1. Blocking — advertised schemas now honour the ids_as_strings contract.** New `crate::schema::widen_id_properties` (keyed variant of the existing `strip_int_formats` tree walk, `mcp`-gated) widens every `ID_KEYS` property to `["integer", "string"]` across the whole schema tree, `$defs` included — scalar `integer`, the `items` of `fragment_ids`, and the `["integer","null"]` of `memory_id` (→ `["integer","string","null"]`). Wired via explicit `input_schema` / `output_schema` expressions on both tools. Input side: `fragments[].id` is widened the same way (string form was accepted but unannounced); `explain_compilation`'s own top-level `fragment_id` parameter deliberately stays `integer`-only (it deserializes as a strict u64), pinned by a dedicated test.

Red → green proof: the 4 new structural tests were written and run BEFORE the widening —
```
test_compile_context_output_schema_advertises_string_ids ....... FAILED
test_explain_compilation_output_schema_advertises_string_ids ... FAILED
test_compile_context_input_schema_advertises_string_fragment_id  FAILED
test_explain_compilation_input_schema_keeps_top_level_fragment_id_strict FAILED
```
then all green after. They assert against the REAL advertised schemas (`McpServer::*_tool_attr()`), collecting every ID-named property recursively — so a future field or `$defs` reshuffle that loses the widening fails loudly. (No `jsonschema` crate in the deps, so structural assertions as suggested.)

**2. CHANGELOG** — the three wave-5 entries moved from `### Fixed` to `### Added` in `[Unreleased]`, with a note about the schema widening added to the ids_as_strings entry.

**3. `deserialize_optional_id`** — now matches over a `serde_json::Value` instead of an untagged enum: a rejected value gets an error naming it and the accepted forms (e.g. `invalid id 1.5 (expected a u64 number or a decimal u64 string)`) instead of serde's opaque "did not match any variant". 4 direct unit tests in `wire_tests.rs`: numeric id (0.8.0 backward-compat), decimal string, `"abc"` rejected, `1.5` rejected — the `1.5` clear-message assertion was RED against the untagged enum, green after the rework.

**Gates re-run, all green:** `cargo fmt --all --check` · workspace clippy pedantic (velesdb-python/node excluded) · `clippy -p velesdb-node --lib` · `RUSTFLAGS=-Dwarnings cargo check -p velesdb-memory --no-default-features --features context` · `cargo test -p velesdb-memory` (187 unit + 154 BDD/property + doctest, 0 failed) · release build + `mcp_e2e.py` OK.

PR #1444, opened 2026-07-18
feat(memory): media source storage + screenshot supersession [EPIC-P-071/US-009]

cyberlife-coder:
## Summary

PR2 of 3 for US-009 (media fragments) of EPIC-P-071 — two pieces per the validated spike design:

**A. Media source storage (memory bridge)**
- `MemoryService::compile_context` (`policy.store_sources`, default on) now persists a media fragment's base64 payload alongside its caption when it stores a compiled source, under a new reserved key `_veles_ctx_source_media` (`{mime, bytes_b64}`).
- Embedding for a media source is a deterministic placeholder derived from the raw bytes' hash — never the text embedder over the (often blank) caption or over the base64 itself — because `retrieve_context_source` resolves media sources by content-addressed hash/slot only, never by vector search (documented in code).
- `MemoryService::retrieve_context_source` now returns `ContextSource { content, media: Option<MediaRef> }` (was `String`); `media` is `#[serde(default)]` so every pre-PR2 text-only source round-trips unchanged.
- PR1's provisional `drop.media_unavailable` verdict is **gone**: an unfit media fragment now externalizes exactly like text (`Retrieve`, `budget.externalize`, resolvable `ctx://source` handle) — the memory bridge can actually resolve it now. A duplicate media fragment whose twin also failed to pack recovers through its own handle too (previously `None`).
- The source storage slot stays keyed on the SAME handle namespace PR1 already established (`stable_id(fragment.content)`, i.e. the caption) — not a second, bytes-addressed namespace — documented in code (`store_context_sources`).
- `MAX_FACT_BYTES` finding from the spike: verified (documented) that `MAX_MEDIA_BYTES`/`MAX_TOTAL_MEDIA_BYTES`, enforced in `compile()`'s `validate_media` before `store_context_sources` ever runs, already bound the new field's size — `MAX_FACT_BYTES` governs the unrelated MCP `remember`/`extract` text ceiling.

**B. Screenshot supersession (classify.rs)**
- New whole-batch pass `classify::screenshot_supersession`, symmetric to `dedup::find_duplicates`: fragments sharing `media` + `kind: "screenshot"` + the same `metadata.target` value form a succession series. Only the LAST one (input order, no clock) stays inline (`Preserve`, budget permitting); every earlier one is reclassified `retrieve.screenshot_superseded` and excluded from packing entirely (regardless of budget), externalized behind a resolvable handle with an explicit reason.
- A screenshot with no `metadata.target` is never superseded (no target is no evidence of succession).
- Opt-outable via `policy.disabled_rules: ["retrieve.screenshot_superseded"]`, like every other named rule.

Also updated: MCP `retrieve_context_source` tool result (`media` field added), Python `retrieve_context_source` binding (now returns a dict, was a bare `str`), CHANGELOG `[Unreleased]`, README "media fragments" section (PR1/3 → PR2/3).

## RED captured

Stashed the implementation (`context.rs`, `classify.rs`, `memory_bridge.rs`, `model.rs`, `mcp/context_tools.rs`) with tests in place: `cargo test -p velesdb-memory --features context,persistence` failed to **compile** —
`cannot find function screenshot_supersession` (6 sites in `classify_tests.rs`) and `struct Analysis<'_> has no field named superseded` (`media_pipeline_tests.rs`) — then restored and re-ran green.

## Gates (all green)

- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets --features persistence,gpu,update-check --exclude velesdb-python --exclude velesdb-node -- -D warnings -D clippy::pedantic`
- `cargo clippy -p velesdb-node --lib -- -D warnings`
- `RUSTFLAGS=-Dwarnings cargo check -p velesdb-memory --no-default-features --features context`
- `cargo test -p velesdb-memory` — all binaries green (263 lib unit tests, 52 compiler BDD, 35 memory-bridge BDD, etc.)
- `python3 crates/velesdb-memory/examples/context_savings/real_measures/mcp_e2e.py` — OK
- `wasm-pack test --node crates/velesdb-wasm` — all green (bridge stays wasm-reachable, no clock, no missing std API)

## Coverage (llvm-cov, touched files)

| File | Line coverage |
|---|---|
| `context.rs` | 98.89% |
| `context/classify.rs` | 99.28% |
| `context/memory_bridge.rs` | 97.73% |
| `context/model.rs` | 100% |
| `mcp/context_tools.rs` | 89.86% |

`mcp/context_tools.rs` sits just under the 90% target — the uncovered lines (37-39, 41, 57-59, 75-77) are pre-existing, untouched schema-generation panic/error branches unrelated to this diff; the added `media` field and its handler path are fully exercised by `test_retrieve_context_source_tool_round_trips_original`.

## Wire-shape choice

`ContextSource { content: String, media: Option<MediaRef> }` extends the existing return type (was a bare `String`) rather than adding a second retrieval method — `media` is `#[serde(default)]`/`skip_serializing_if` so every existing JSON consumer (MCP) and the BDD goldens with no media fragments are byte-identical.

## Motivated deviations

- Kept the storage slot content-addressed on the caption hash (PR1's existing `provenance::handle_for` scheme), not on the media bytes hash, to avoid a second handle namespace — documented the caption-collision edge case (two media fragments sharing an identical, often-blank caption collide onto one slot, first-write-wins) as inherited, unchanged PR1 behavior, not a new PR2 bug.
- Did not add a runtime `MAX_FACT_BYTES`-style guard inside `store_context_sources`; `validate_media` in `compile()` already bounds every fragment before this code path ever runs (documented instead of duplicating validation).

## Test plan

- [x] `cargo test -p velesdb-memory` (all features)
- [x] `wasm-pack test --node crates/velesdb-wasm`
- [x] `mcp_e2e.py`
- [x] Coverage via `cargo llvm-cov`

### DIFF
--- a/crates/velesdb-memory/CHANGELOG.md
+++ b/crates/velesdb-memory/CHANGELOG.md
@@ -11,6 +11,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Media source storage & screenshot supersession (experimental, PR2/3 of
+  US-009 in EPIC-P-071)** — the memory bridge now persists a media
+  fragment's base64 payload alongside its caption when storing a compiled
+  source (reserved key `_veles_ctx_source_media`, embedded with a
+  deterministic bytes-hash-derived placeholder vector rather than the text
+  embedder — `retrieve_context_source` resolves media sources by
+  content-addressed hash only, never by vector search). PR1's provisional
+  `drop.media_unavailable` verdict is gone: a media fragment that cannot fit
+  the budget now externalizes exactly like text (`budget.externalize`, a
+  resolvable `ctx://source` handle), and a duplicate media fragment whose
+  twin also failed to pack recovers through its own handle too.
+  `MemoryService::retrieve_context_source` returns the new `ContextSource {
+  content, media? }` shape (`media` is `#[serde(default)]`, so every
+  pre-PR2 text-only source round-trips unchanged); the MCP
+  `retrieve_context_source` tool result gained the same optional `media`
+  field, and the Python `retrieve_context_source` binding now returns a
+  dict instead of a bare string.
+  **Screenshot supersession**: fragments sharing `media` + `kind:
+  "screenshot"` + the same `metadata.target` value are a succession
+  series — only the LAST one (input order, no clock) stays inline
+  (`Preserve`, budget permitting); every earlier one is proactively
+  reclassified `retrieve.screenshot_superseded` and externalized behind a
+  resolvable handle, regardless of budget, with an explicit reason. A
+  screenshot with no `metadata.target` is never superseded (no target is no
+  evidence of succession). Opt out per request via
+  `policy.disabled_rules: ["retrieve.screenshot_superseded"]`. Byte-compat:
+  a request with no media is unaffected.
 - **Media fragments (experimental, PR1/3 of US-009 in EPIC-P-071)** —
   `ContextFragment.media: Option<MediaRef>` lets a fragment carry an inline
   base64-encoded image (`mime` + `bytes_b64`) alongside its text/caption
@@ -21,12 +48,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `ceil(width * height / 750)`; unsupported mimes or unreadable headers fall
   back to a safe text-based over-count). Capped at 4 MiB of base64
   (`limits::MAX_MEDIA_BYTES`), separate from the existing per-fragment text
-  cap; malformed base64 is rejected at validation time. PR1 ships no binary
-  retrieval store: a media fragment that cannot fit the budget is `drop`ped
-  with an explicit reason (`drop.media_unavailable`) rather than handed a
-  `ctx://source` handle that would not resolve — externalized-media storage,
-  screenshot expiry, and the Node/WASM/wire surface land in a later PR. Wire-
-  compatible: `media` is `#[serde(default)]`, so every existing request still
+  cap; malformed base64 is rejected at validation time. Wire-compatible:
+  `media` is `#[serde(default)]`, so every existing request still
   deserializes unchanged.
 
 ## [0.6.0] - 2026-07-06
--- a/crates/velesdb-memory/CHANGELOG.md
+++ b/crates/velesdb-memory/CHANGELOG.md
@@ -17,7 +17,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   source (reserved key `_veles_ctx_source_media`, embedded with a
   deterministic bytes-hash-derived placeholder vector rather than the text
   embedder — `retrieve_context_source` resolves media sources by
-  content-addressed hash only, never by vector search). PR1's provisional
+  content-addressed hash only, never by vector search). A media fragment's
+  handle — and its storage slot, still under the same salted system-fact
+  namespace — is keyed on the **raw decoded bytes' hash** (the identity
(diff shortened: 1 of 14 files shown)


cyberlife-coder:
## Review addressed — bytes-identity fix pushed

**Bloquant (collision de handles média) — corrigé, TDD rouge→vert :**

**RED capturé** (commit `18d82e6a`, tests seuls, code inchangé) — les 3 probes épinglées comme tests BDD échouent exactement sur le bug :
- P1 `test_two_blank_caption_media_fragments_with_different_bytes_get_distinct_resolving_handles` — `assertion left != right failed: different bytes must never share a handle` (les deux handles = `ctx://source/14695981039346656037` = `stable_id("")`)
- P2 `test_cross_compile_blank_caption_media_sources_never_serve_stale_bytes` — `an occupied slot from an earlier compile must never serve stale bytes` (bytes de A servis pour le handle de B)
- P3 `test_superseded_blank_caption_screenshots_resolve_their_own_bytes_not_the_survivors` — handles superseded identiques, bytes du survivant servis
- Fixture corrigée : `distinct_media_b64(seed)` prend un seed explicite **indépendant de la caption** (le couplage rendait le scénario du bug inexprimable) + fold multiplicatif (les seeds de même longueur ne collisionnent plus par somme).

**GREEN** (commit `860e97d2`) — identité média = bytes :
- `Analysis::handle_hash()` : `raw_hash` (bytes décodés, l'identité de la dédup PR1) pour un fragment média, `content_hash` sinon — utilisé à TOUS les points de mint (verdicts retrieve/dup/supersession, `retrieval_handles`, `out.sources`).
- Bridge : `fragment_handle_hash` (jumeau documenté en lockstep), map de stockage keyée sur la même identité, **première occurrence gagne** (cohérent avec le twin gardé par la dédup) ; le slot reste dans l'espace salté `SOURCE_ID_SALT` — pas de nouveau namespace. Bonus : plus de re-décodage pour l'embedding placeholder (le hash du handle EST le raw_hash).
- Conséquence assumée documentée : bytes identiques + captions différentes ⇒ même handle, résout la caption de l'instance gardée (« la caption divergente ne survit pas », sémantique PR1). Handles texte byte-identiques à l'existant.

**Mineurs :**
1. ✅ `test_byte_identical_screenshot_duplicate_wins_over_supersession_and_resolves` — 2 screenshots byte-identiques + 1 plus récent, même target : #0 `retrieve.screenshot_superseded`, #1 `drop.duplicate` (l'arm dup gagne), #2 `Preserve`, et le handle du dup résout ses bytes.
2. ✅ `test_stored_media_sources_are_invisible_to_normal_recall` — le fact média (placeholder one-hot, hub-marqué) ne surface jamais au `recall`.
3. ✅ CHANGELOG : ligne croissance stockage (fact média jusqu'à 4 MiB, au-delà du plafond `MAX_FACT_BYTES` qui ne couvre que l'entrée texte MCP ; borné par `MAX_MEDIA_BYTES`/`MAX_TOTAL_MEDIA_BYTES` + `source_ttl_seconds`) + note identité bytes.

**Gates (tous verts) :** `cargo fmt --check` · clippy workspace pedantic (`--features persistence,gpu,update-check`, excl. python/node) · `clippy -p velesdb-node --lib` · `RUSTFLAGS=-Dwarnings cargo check --no-default-features --features context` · `cargo test -p velesdb-memory` (13 binaires, 40 tests bridge BDD dont les 5 nouveaux) · `mcp_e2e.py` OK · `wasm-pack test --node` OK.

**Couverture (llvm-cov, fichiers touchés) :** context.rs 98.91 % · classify.rs 99.28 % · memory_bridge.rs 97.77 % · provenance.rs 100 % · model.rs 100 %.

PR #1457, opened 2026-07-19
fix(graph): reject add_edge to non-existent nodes (#1442)

### LINKED ISSUE #1442: REST: nodes created implicitly via addEdge are invisible to graph-nodes / MATCH
Found while verifying a documented REST example (PR #1441): creating edges via the REST relations path registers the edges, but the nodes they reference — when never explicitly created — do not appear through the graph-nodes endpoint, and a subsequent `MATCH` over that collection returns empty.

Repro sketch (TypeScript SDK, REST backend): createCollection → addEdge(a→b) without explicit node creation → MATCH over the collection → empty result, while the same implicit-node pattern works through the in-process memory wedge (`relate` builds traversable graphs).

Expected: either implicit nodes are surfaced consistently (parity with the wedge/graph engine semantics), or the REST contract documents that MATCH requires explicitly created nodes.

Scope: velesdb-server REST layer vs core graph semantics. Discovered during EPIC-P-071 wave-5 doc verification; not blocking that PR (the documented example was switched to a fully verified VelesQL vector path).

cyberlife-coder:
## Description

Fixes #1442 — nodes created implicitly via `addEdge`/`INSERT EDGE` were invisible to `graph-nodes`/MATCH on schemaless graph collections (the SDK default). Root cause: `Collection::add_edge`/`add_edges_batch` never validated that an edge's `source`/`target` had a stored payload unless the collection used a *strict* graph schema — silently accepting edges to node IDs that were never created. `GET /graph/nodes`, `all_node_ids()`, and `MATCH` all resolve their node set from the payload store, not the edge store, so those "phantom" edges were permanently invisible outside direct edge queries.

The investigation showed the in-process memory wedge (`velesdb-memory`'s `relate()`) already required both endpoints to exist via its own client-side check — core just didn't enforce the same invariant. The fix makes `add_edge`/`add_edges_batch` reject (not silently accept, not auto-create) an edge to a missing endpoint with `Error::NodeNotFound` (`VELES-022`, REST `404`), matching what strict-schema collections and the wedge already did.

An adversarial multi-angle review of the initial fix (6 independent finder passes: line-by-line diff scan, removed-behavior audit, cross-file caller trace, reuse/simplification, efficiency/hotspots, altitude/conventions) surfaced four additional real bugs beyond the original scope, all fixed in the second commit:

- **`POST /collections/{name}/relations`** (the exact REST endpoint the original bug report was about) mapped every non-`EdgeExists` `add_edge` error to a hardcoded `500`, so the new `NodeNotFound` surfaced as *Internal Server Error* instead of `404`/`VELES-022`. This endpoint had **zero** REST test coverage before this PR.
- **`velesdb-memory`'s `relate()`** downgraded the identical concurrent-delete race to a generic internal error instead of its own documented `NotFound`/`invalid_params` contract.
- **`velesdb-migrate`'s FK-relations phase** writes edges into a fresh, edge-only `GraphCollection` that never otherwise gets node payloads — under the naive fix, *every* migrated edge would now be rejected outright. It now stubs each endpoint before inserting (deduplicated across the whole relation), and a batch-level validation failure degrades to per-edge accounting instead of aborting the entire migration. This path also had zero test coverage before this PR.
- **`tauri-plugin-velesdb`**'s `add_edge`/`add_edges_batch` collapsed every core error into a generic `InvalidConfig` instead of using the crate's own `#[from]` conversion that preserves the VELES-XXX code. ⚠️ **Not compile-verified** — this crate needs GTK/webkit2gtk system libraries unavailable in the sandbox this PR was authored in; please confirm in CI.

Also from review: the new validation helper now reuses the existing `get_node_payload` instead of re-implementing the lock+retrieve; two BFS/DFS stress tests were switched from per-edge to bulk node seeding (a real fsync hotspot the naive fix introduced); added schemaless-mode test coverage for the mixed-endpoint case (only source or only target exists); documented the pre-existing check-then-write race shared by both the strict and schemaless validation paths (not newly introduced, not fixed — closing it fully would need a lock-scope change with real deadlock risk I didn't want to rush); de-duplicated a 3×-copy-pasted node-seeding block in `velesdb-cli`'s `e2e_complete.rs`.

`velesdb-migrate`, `velesdb-mobile`, `velesdb-wasm`, and (beyond the one contract fix above) `velesdb-memory` needed no other changes: migrate's non-relations paths never exercised dangling-edge paths, mobile/wasm ship their own self-contained graph stores not backed by `velesdb_core::Collection`.

## Type of Change

- [x] 💥 Breaking change — schemaless graph collections (the default) now reject `add_edge`/`add_edges_batch` calls whose endpoints weren't created first, where they previously silently succeeded. See the CHANGELOG migration note.
- [x] 📚 Documentation update (`GRAPH_PATTERNS.md`, `VELESQL_SPEC.md`, OpenAPI spec, CHANGELOG)

## How Has This Been Tested?

- [x] Unit tests — `velesdb-core` (5271 passed, 2 new), `velesdb-server` (152 lib + 87 `api_integration` incl. 2 new relations tests), `velesdb-cli` (all binaries green), `velesdb-migrate` (257 passed incl. 1 new e2e test proving the graph-relations phase survives the new invariant), `velesdb-memory` (267 passed)
- [x] Integration tests — full BDD suite (632 passed), all `velesdb-server` integration binaries
- [ ] Manual testing — not performed; relied on the above plus `cargo clippy --all-targets -D warnings -D clippy::pedantic` (clean on every crate compilable in this sandbox) and `cargo fmt --all -- --check` (clean)

**Test Configuration:**
- OS: Linux (sandboxed container)
- Rust version: matches workspace toolchain

## Checklist

- [x] My code follows the style guidelines of this project (clippy pedantic + rustfmt clean)
- [x] I have performed a self-review of my code (adversarial multi-angle review, see above)

## Additional Notes

`tauri-plugin-velesdb` could not be built or tested in the authoring sandbox (missing `atk`/`gtk`/`webkit2gtk` system libraries, and the package mirror needed to install them was unreachable). The one change there (`crates/tauri-plugin-velesdb/src/commands_graph.rs`) is a small, type-checked-by-inspection swap from a manual `.map_err` to `?` relying on an existing `#[from] velesdb_core::Error` impl — please have CI confirm it compiles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`velesdb-core` / `velesdb-server`**: `add_edge` and `add_edges_batch`
+  now reject an edge whose `source` or `target` node has no stored payload
+  (#1442), instead of silently accepting it. Previously, on schemaless
+  graph collections (the default), an edge could reference a node ID that
+  was never explicitly created — the edge was stored, but the node stayed
+  invisible to `GET /graph/nodes`, `all_node_ids()`, and `MATCH`, since all
+  three resolve their node set from the payload store, not the edge store.
+  The REST layer now returns `404 NodeNotFound` (`VELES-022`) for such an
+  edge, matching the existing strict-schema behavior and the
+  `velesdb-memory` wedge's `relate()`, which already required both
+  endpoints to exist. **Migration note:** code that called `addEdge`/
+  `INSERT EDGE` without first creating the endpoint nodes (via a point
+  upsert or `PUT /graph/nodes/{id}/payload`) must now create them first —
+  see `docs/guides/GRAPH_PATTERNS.md`.
+
 - **`velesdb-memory`**: hardened the MCP server against a leaked client
   process (#1448). The server itself was already healthy (it exits cleanly
   on stdin EOF), but a client that leaks its child process — observed in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   endpoints to exist. **Migration note:** code that called `addEdge`/
   `INSERT EDGE` without first creating the endpoint nodes (via a point
   upsert or `PUT /graph/nodes/{id}/payload`) must now create them first —
-  see `docs/guides/GRAPH_PATTERNS.md`.
+  see `docs/guides/GRAPH_PATTERNS.md`. Follow-on fixes found in review:
+  `POST /collections/{name}/relations` (the point-relations REST endpoint)
+  previously mapped this same error to a generic `500`; it now returns
+  `404`/`VELES-022` like every other graph endpoint. `velesdb-memory`'s
+  `relate()` previously downgraded the identical concurrent-delete race to
+  a generic internal error instead of its documented `NotFound` contract;
+  fixed. `velesdb-migrate`'s FK-relations phase writes into a fresh,
+  edge-only graph collection that never otherwise gets node payloads —
+  it now stubs each edge endpoint before inserting, and a batch-level
+  validation failure degrades to per-edge accounting instead of aborting
+  the whole migration.
 
 - **`velesdb-memory`**: hardened the MCP server against a leaked client
   process (#1448). The server itself was already healthy (it exits cleanly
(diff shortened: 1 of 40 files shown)


cyberlife-coder:

cyberlife-coder:
Closing in favor of #1460 — same content, recreated on a branch matching this repo's Git Flow naming policy (`fix/*`) so the governance check passes. The `claude/admiring-fermi-fxrsdv` branch name doesn't match any of the allowed prefixes for targeting `develop`.

PR #1460, opened 2026-07-19
fix(graph): reject add_edge to non-existent nodes (#1442)

### LINKED ISSUE #1442: REST: nodes created implicitly via addEdge are invisible to graph-nodes / MATCH

cyberlife-coder:
Supersedes #1457, recreated on a branch matching this repo's Git Flow naming policy (`fix/*`) so the governance check passes. Same original commits, plus a re-fix pass (see below) after adversarial review found a residual race.

## Description

Fixes #1442 — nodes created implicitly via `addEdge`/`INSERT EDGE` were invisible to `graph-nodes`/MATCH on schemaless graph collections (the SDK default). Root cause: `Collection::add_edge`/`add_edges_batch` never validated that an edge's `source`/`target` had a stored payload unless the collection used a *strict* graph schema — silently accepting edges to node IDs that were never created. `GET /graph/nodes`, `all_node_ids()`, and `MATCH` all resolve their node set from the payload store, not the edge store, so those "phantom" edges were permanently invisible outside direct edge queries.

The fix makes `add_edge`/`add_edges_batch` reject (not silently accept, not auto-create) an edge to a missing endpoint with `Error::NodeNotFound` (`VELES-022`, REST `404`), matching what strict-schema collections and the `velesdb-memory` wedge already did. An earlier adversarial review (6 finder passes) found four additional real bugs beyond the original scope — REST `relations` 500, `velesdb-memory`'s `relate()` race downgrade, `velesdb-migrate`'s FK-relations breakage, `tauri-plugin-velesdb`'s error collapsing — all fixed in the original commit set (`4038ead8`, `51d7377e`, `101338dd`, `7d7bd3ab`).

## Re-fix: two additional bugs closed in this pass

A second adversarial review of the fix itself found two more issues, both closed here with TDD (red tests committed before the fix):

### 1. Concurrent-delete phantom-edge race (core)

`add_edge`/`add_edges_batch` validated endpoint existence under `payload_storage.read()`, then **released that guard** before writing the edge (WAL append + edge-store apply). A concurrent `delete()` of an endpoint could land in that window and still leave a "phantom" edge — present in the edge store, invisible to `all_node_ids()`/MATCH, since both resolve nodes from the payload store.

**Fix**: both entry points now hold `payload_storage`'s read guard from the referential-integrity check through the end of the write; `delete()` needs the **write** guard to remove a payload, so it blocks until the edge is fully durable, at which point `delete()`'s own edge cascade sees and removes it. No compensation logic, no rollback — for the batch path the same guard covers the whole batch, so "an endpoint disappears mid-batch" is impossible by construction. Full mechanism documented as an additive "Collection-level lock order" section in `docs/CONCURRENCY_MODEL.md` (position **3b** for `edge_wal_lock`, nested inside `payload_storage`(3) for this path — the frozen `LockRank` table above it is untouched).

Validator signatures (`validate_edge_endpoints_exist`, `validate_edge_referential_integrity`, `endpoint_node_type`) now take the caller's already-acquired `&LogPayloadStorage` instead of re-locking internally — re-locking from inside the outer guard's scope would risk a self-deadlock once a writer is queued on the same `parking_lot` `RwLock`.

**Residual cost**: payload writers can stall behind an in-flight edge write for up to one `fsync` (typically 0.05-5 ms on SSD) — the same per-edge fsync cost that already existed. Documented in `docs/guides/WRITE_CONCURRENCY.md` with a batch-size recommendation (≤10k edges).

### 2. Migrate seed-retry bug (velesdb-migrate)

`velesdb-migrate`'s FK-relations seeding phase marked a node id as "seeded" (`seeded_nodes.insert(id)`) **before** checking whether its stub `upsert_node_payload` call actually succeeded. A failed upsert permanently skipped any retry for that node later in the same relation, silently orphaning every edge referencing it. Now only marks a node seeded on `Ok`; a failure logs a warning and leaves the id retryable (upserts are idempotent, so a retry is always safe).

## Red tests (TDD, archived)

```
test collection::core::graph_api_tests::tests::add_edge_concurrent_delete_never_leaves_phantom_edge ...
thread '...' panicked at crates/velesdb-core/src/collection/core/graph_api_tests.rs:1267:13:
iteration 0: delete() must block on payload_storage while add_edge holds its read guard through the WAL pause
FAILED
test collection::core::graph_api_tests::tests::add_edges_batch_concurrent_delete_never_leaves_phantom_edge ...
thread '...' panicked at crates/velesdb-core/src/collection/core/graph_api_tests.rs:1285:13:
iteration 0: delete() must block on payload_storage while add_edges_batch holds its read guard through the WAL pause
FAILED
test collection::core::graph_api_tests::tests::stress_add_edge_concurrent_delete_never_leaves_phantom_edge ...
thread '...' panicked at crates/velesdb-core/src/collection/core/graph_api_tests.rs:1330:13:
iteration 25: phantom edge under stress
FAILED
```

```
test pipeline_graph::tests::seed_retries_failed_node_on_next_occurrence ...
thread '...' panicked at crates/velesdb-migrate/src/pipeline_graph.rs:482:9:
a failed upsert must not mark the node as seeded
FAILED
```

All four green after the corresponding fix commit.

## Review points addressed

- **Race closed by holding the guard, not compensating.** A compensating approach (write the edge, then check-and-undo if the endpoint vanished) was considered and rejected: it leaves a crash window where the ADD is fsynced but a crash hits before the compensating REMOVE, and WAL replay bypasses referential-integrity validation by design (see "Known limitation" below) — so a crash-interrupted compensation would recreate the phantom permanently on the next open. Holding the guard closes the window instead of racing to patch it after the fact, and needs no rollback path.
- **Seed-retry**: fixed as described above; covered by an injectable-upsert unit test since simulating a real storage failure end-to-end isn't practical.

## Known, accepted limitations (tracked)

- Edges loaded from a pre-existing WAL/snapshot at `Collection::open` (replay) are trusted as-is and never re-validated — replay intentionally bypasses referential-integrity validation, since a database predating the original #1442 fix may legitimately be edge-only. A database created before that fix may still contain legacy phantom edges. Tracked in #1469 (`velesdb-cli graph doctor <collection> [--purge|--stub]`).
- The strict-vs-schemaless error-type divergence (`SchemaValidation` vs. `NodeNotFound`) from the original fix remains deliberate (changing it now would be breaking). Tracked for the next major version in #1470.

## Gates (all local, this pass)

| Gate | Result |
|---|---|
| `cargo fmt --all --check` | pass |
| `cargo clippy --workspace --all-targets --features persistence,gpu,update-check --exclude velesdb-python --exclude velesdb-node -- -D warnings -D clippy::pedantic` | pass |
| `cargo clippy -p velesdb-node --lib` | pass |
| `cargo test -p velesdb-core --features persistence -- --test-threads=1` | pass — 5287 passed, 0 failed |
| `cargo test -p velesdb-migrate` | pass — 33 passed, 0 failed |
| `cargo test -p velesdb-server` | pass — 0 failed |
| `cargo check --no-default-features` | **fails identically on unmodified `origin/develop`** (pre-existing `velesdb-server` breakage unrelated to this PR's files — verified in a separate baseline worktree) |
| `wasm-pack test --node crates/velesdb-wasm` | pass — 29 passed, 0 failed |
| `RUSTFLAGS="--cfg loom" cargo +nightly test --features loom,persistence -p velesdb-core --test loom_tests` | pass — 5 passed, including the new `loom_payload_guard_excludes_delete_during_edge_write` model |
| grep: no `get_node_payload`/nested `payload_storage.read()` call under an already-held guard | clean |
| grep: no `payload_storage` acquisition while holding `edge_wal_lock` | clean |
| grep: no AI attribution in any commit on this branch | clean |

Follow-up issues: #1469, #1470.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   supersession target could both drop from compiled context instead of one
   surviving. Media dedup now re-anchors onto the freshest non-superseded
   occurrence. (#1453)
+- **`velesdb-core` / `velesdb-server`**: `add_edge` and `add_edges_batch`
+  now reject an edge whose `source` or `target` node has no stored payload
+  (#1442), instead of silently accepting it. Previously, on schemaless
+  graph collections (the default), an edge could reference a node ID that
+  was never explicitly created — the edge was stored, but the node stayed
+  invisible to `GET /graph/nodes`, `all_node_ids()`, and `MATCH`, since all
+  three resolve their node set from the payload store, not the edge store.
+  The REST layer now returns `404 NodeNotFound` (`VELES-022`) for such an
+  edge, matching the existing strict-schema behavior and the
+  `velesdb-memory` wedge's `relate()`, which already required both
+  endpoints to exist. **Migration note:** code that called `addEdge`/
+  `INSERT EDGE` without first creating the endpoint nodes (via a point
+  upsert or `PUT /graph/nodes/{id}/payload`) must now create them first —
+  see `docs/guides/GRAPH_PATTERNS.md`.
+
 - **`velesdb-memory`**: hardened the MCP server against a leaked client
   process (#1448). The server itself was already healthy (it exits cleanly
   on stdin EOF), but a client that leaks its child process — observed in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   it now stubs each edge endpoint before inserting, and a batch-level
   validation failure degrades to per-edge accounting instead of aborting
   the whole migration.
+- **`velesdb-core`**: closed a race left by the #1442 fix above —
+  `add_edge`/`add_edges_batch` used to release the payload-store guard
+  right after checking that both endpoints exist, then separately acquire
+  the edge-WAL lock to write the edge. A concurrent `delete()` of an
+  endpoint could land in that window and still leave a "phantom" edge
+  (present in the edge store, invisible to `all_node_ids()`/MATCH). Both
+  entry points now hold the payload-store read guard from the check
+  through the end of the write, so a racing `delete()` blocks until the
+  edge is durable instead of running ahead; see the "Collection-level lock
+  order" section of `docs/CONCURRENCY_MODEL.md` for the full mechanism.
+  Also fixed: `velesdb-migrate`'s FK-relations node-seeding marked a node
+  as seeded even when its stub upsert failed, permanently skipping any
+  retry for that node later in the same relation — it now only marks a
+  node seeded on a successful upsert. **Known, accepted limitation**:
+  edges loaded from a pre-existing WAL/snapshot (replay) are not
+  re-validated, so a database created before the original #1442 fix may
+  still contain legacy phantom edges; tracked in
+  [#1469](https://github.com/cyberlife-coder/VelesDB/issues/1469) (a
+  `velesdb-cli graph doctor` audit/repair subcommand). The
+  strict-vs-schemaless error-type divergence (`SchemaValidation` vs.
+  `NodeNotFound`) noted above remains deliberate; unifying it is tracked
+  for the next major version in
+  [#1470](https://github.com/cyberlife-coder/VelesDB/issues/1470).
 
 - **`velesdb-memory`**: hardened the MCP server against a leaked client
   process (#1448). The server itself was already healthy (it exits cleanly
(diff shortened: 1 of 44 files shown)


cyberlife-coder:

cyberlife-coder:

PR #1524, opened 2026-07-21
feat(memory): streamable-HTTP transport + automated multi-client daemon installer

cyberlife-coder:
## Summary
- **Part 1 — HTTP transport**: adds a non-default `http` feature to `velesdb-memory`. `--http`/`--http-port` (or `VELESDB_MEMORY_HTTP`/`VELESDB_MEMORY_HTTP_BIND`) serve the MCP server over rmcp's streamable-HTTP transport instead of stdio, so several MCP clients (Claude Code, Claude Desktop, Windsurf, …) can share ONE daemon process instead of each spawning its own stdio process and fighting over the store's single-writer `flock`. The hash/ollama embedder choice stays a pure `VELESDB_MEMORY_EMBEDDER` runtime switch either way — never a build-time one. New `/health` liveness endpoint. Default binary is unchanged: stdio-only, tiny, offline.
- **Part 2 — installer**: `scripts/install-memory-daemon.sh` builds the daemon (`--features ollama,http`, always both regardless of the runtime embedder choice), runs it as a macOS `launchd` agent, and wires Claude Code / Claude Desktop / Windsurf to it. `--uninstall` reverses everything without ever touching the store. Non-macOS: builds + wires clients, skips the daemon step with a clear note.

## TDD (RED → GREEN)
Commit `8fc9ff11` is the RED commit: `tests/http_transport.rs` was written against a `velesdb_memory::http::router` that didn't exist yet, gated via a new `[[test]] required-features = ["http"]` Cargo.toml entry (so it doesn't break default builds). Captured red:

```
$ cargo test -p velesdb-memory --features http --no-run
error: the package 'velesdb-memory' does not contain this feature: http
help: there is a similarly named feature: mcp
```

Commit `97560778` is GREEN: adds the `http` feature, `src/http.rs` (the router), and wires `main.rs`. All 4 `http_transport` tests + the adversarial `http_lock_contention` test pass.

## Gates
| Gate | Result |
|---|---|
| `cargo fmt --all --check` | PASS |
| `cargo clippy -p velesdb-memory --all-targets --features http,ollama -- -D warnings` | PASS. (Was blocked on a pre-existing, unrelated `clippy::similar_names` failure in `examples/locomo/dump.rs:229`, introduced in `efd3aea6` #1301 long before this branch — fixed upstream on `develop` by #1525 and picked up by this branch's rebase onto it.) |
| `cargo test -p velesdb-memory --features http` | PASS — 347 (default-feature unit tests, unaffected) + all http-gated suites, including the new ones |
| `cargo test -p velesdb-memory` (default features, no `http`) | PASS — confirms zero regression for the default stdio-only build |
| `cargo check --workspace` | PASS |
| `shellcheck scripts/install-memory-daemon.sh` | PASS (clean, after fixing one SC2034 unused-var warning) |

## Deviations from the plan (with justification)
1. Concurrency tests use one HTTP connection per task, not one shared session. The plan didn't mandate either; sharing ONE `RunningService`/session across 20 concurrent `call_tool` invocations reproducibly hung (confirmed via `sample` stack traces: both concurrency tests stuck in `mio::…::select`, zero CPU). Per-task-own-connection is both more representative of the real scenario this feature targets (independent MCP client processes, each with their own session) and avoids relying on undocumented concurrent-call-multiplexing semantics of rmcp's client-side streamable-HTTP transport that aren't exercised anywhere in rmcp's own upstream test suite either.
2. `#[tokio::test(flavor = "multi_thread", worker_threads = 8)]` on all 4 `http_transport` tests, matching the real production runtime (`tokio::runtime::Runtime::new()` in `main.rs` is multi-thread by default) more faithfully than the macro's single-thread default. **Correction (2026-07-22): the single-thread test runtime was NOT the root cause of the earlier hang**, despite what an earlier revision of this description claimed. The hang was a real, mechanically-confirmed deadlock in `velesdb-core`'s `Collection` locking (flat CPU time over 25s, not starvation; zero rmcp/axum/tower frames in the stuck call chain) — two independent root causes: rayon pool exhaustion in `crates/velesdb-core/src/collection/core/crud.rs`, and an ABBA lock-order inversion across `get_raw` and 7 other `Collection` read-path sites. Both fixed on `develop` via #1527. Full investigation trail (repro, stack samples) archived at `.investigation/http-deadlock-2026-07-22/`. This PR's own concurrency tests (`twenty_concurrent_remembers_all_succeed_with_unique_ids`, `concurrent_remember_and_recall_do_not_deadlock_and_all_facts_recallable`) hung before the core fix landed; re-run 20x each with a hard per-run timeout after rebasing onto `develop` (which now includes #1527): 40/40 passed, zero hangs, zero orphan processes left behind.
3. ~~Pre-existing clippy failure in `examples/locomo/dump.rs` left as-is~~ — resolved upstream (#1525, see Gates table); no longer a deviation.
4. **Added during the 2026-07-22 OOM audit** (not in the original plan, follow-up from the deadlock investigation): `router()` now bounds the two things rmcp's own defaults leave unbounded on `/mcp` — request body size (`tower_http::limit::RequestBodyLimit`, since rmcp's `StreamableHttpService` reads the body directly via `Body::collect`, bypassing axum's extractor-based `DefaultBodyLimit` entirely) and concurrent session count (`BoundedSessionManager`, wrapping `LocalSessionManager`, whose session map otherwise grows without bound). Both configurable (`VELESDB_MEMORY_HTTP_MAX_BODY_BYTES`, `VELESDB_MEMORY_HTTP_MAX_SESSIONS`), both covered by new adversarial tests. Per-fact size was already bounded independent of transport (`crate::limits::MAX_FACT_BYTES`, enforced in the tool handlers) — not touched. Full audit notes, including what was checked and deliberately left alone (session-manager's own zombie-session gap, upstream in `rmcp`; the installer's launchd plist has no memory ulimit, consistent with every other daemon in this repo — none has one either), are in the PR review thread.

## Blockers
None. No permission was refused for any step — `launchctl`, `~/Library/LaunchAgents`, and `cargo install` were never invoked for real (the installer's daemon step and full end-to-end client wiring need a real macOS run outside this sandboxed environment to verify beyond static/syntax checks). What WAS verified in-sandbox: `bash -n`, `shellcheck`, `--help`, a real `--uninstall` dry run (confirmed no-op via `claude mcp list` — no prior `velesdb-memory` user-scope entry existed), and the `jq` filters exercised standalone against mock Desktop/Windsurf config files (add + del round-trip, existing keys preserved).

## Test plan
- [x] `cargo test -p velesdb-memory --features http` — all pass
- [x] `cargo test -p velesdb-memory` (default features) — all pass, no regression
- [x] `cargo fmt --all --check`
- [x] `shellcheck scripts/install-memory-daemon.sh`
- [ ] Manual: run `./scripts/install-memory-daemon.sh` on a real macOS machine and confirm the launchd agent + all three client wirings end to end (not exercised in this sandbox)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4580,18 +4580,28 @@ checksum = "14db48ee17a9ba61810ab1a9c1beb7d06d8136ae39ac25a1137f10d357af01af"
 dependencies = [
  "async-trait",
  "base64 0.22.1",
+ "bytes",
  "chrono",
  "futures",
+ "http",
+ "http-body",
+ "http-body-util",
  "pastey",
  "pin-project-lite",
+ "rand 0.10.2",
+ "reqwest",
  "rmcp-macros",
  "schemars 1.2.1",
  "serde",
  "serde_json",
+ "sse-stream",
  "thiserror 2.0.19",
  "tokio",
+ "tokio-stream",
  "tokio-util",
+ "tower-service",
  "tracing",
+ "uuid",
 ]
 
 [[package]]
@@ -5394,6 +5404,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d9571ea910ebd84c86af4b3ed27f9dbdc6ad06f17c5f96146b2b671e2976744f"
 dependencies = [
  "bitflags 2.11.0",
+]
+
+[[package]]
+name = "sse-stream"
+version = "0.2.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "39f24a9b78c40b90817bbcd1821c74ddfd74916aadd29403d001532a9195532d"
+dependencies = [
+ "bytes",
+ "futures-util",
+ "http-body",
+ "http-body-util",
+ "pin-project-lite",
 ]
 
 [[package]]
@@ -6768,6 +6791,7 @@ dependencies = [
 name = "velesdb-memory"
 version = "0.10.1"
 dependencies = [
+ "axum",
  "criterion",
  "rmcp",
  "schemars 1.2.1",
@@ -6776,6 +6800,7 @@ dependencies = [
  "tempfile",
  "thiserror 2.0.19",
  "tokio",
+ "tokio-util",
  "ureq",
  "velesdb-core",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6237,6 +6237,7 @@ dependencies = [
  "bytes",
  "http",
  "http-body",
+ "http-body-util",
  "percent-encoding",
  "pin-project-lite",
  "tower-layer",
@@ -6793,6 +6794,7 @@ version = "0.10.1"
 dependencies = [
  "axum",
  "criterion",
+ "futures",
  "rmcp",
  "schemars 1.2.1",
  "serde",
@@ -6801,6 +6803,7 @@ dependencies = [
  "thiserror 2.0.19",
  "tokio",
  "tokio-util",
+ "tower-http 0.7.0",
  "ureq",
  "velesdb-core",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ authors = ["Julien Lange <contact@wiscale.fr>"]
 [workspace.dependencies]
 # Async runtime
 tokio = { version = "1.52", features = ["full"] }
+tokio-util = "0.7"
 
 # Serialization
 serde = { version = "1.0", features = ["derive"] }
--- a/crates/velesdb-memory/Cargo.toml
+++ b/crates/velesdb-memory/Cargo.toml
@@ -68,6 +68,11 @@ tempfile = "3"
 serde_json = { workspace = true }
 tokio = { workspace = true }
 criterion = { workspace = true }
+# MCP client-side streamable-HTTP transport, dev-only: drives the `http`
+# feature's server transport in tests/http_transport.rs with a real client
+# (the same one exercised by rmcp's own upstream test suite) rather than
+# hand-rolled JSON-RPC-over-HTTP requests.
+rmcp = { version = "2.0.0", features = ["client", "transport-streamable-http-client-reqwest"] }
 
 [[bin]]
 name = "velesdb-memory"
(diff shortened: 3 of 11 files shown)


cyberlife-coder:
Pushed `27edddb`: `requested_http_bind` now refuses to bind the HTTP transport to a non-loopback address unless `VELESDB_MEMORY_HTTP_ALLOW_REMOTE=1` is explicitly set. The router has no authentication, so an overridable `VELESDB_MEMORY_HTTP_BIND` host with no restriction (e.g. `0.0.0.0`) would have exposed full read/write store access with no auth and no warning — this closes that gap while keeping the documented loopback-only default working unchanged. Added unit tests for the new `is_loopback_host` check; existing `http_transport`/`http_lock_contention`/`cli_flags` suites still pass unmodified. README updated next to the other HTTP transport flags.

cyberlife-coder:
Codacy still showed `1 high` after the loopback-bind fix, and #1526 (which touches the actual `quick-xml`/`plist` advisories) reports `0 issues` — so the finding isn't a dependency CVE. Most likely culprit: `scripts/install-memory-daemon.sh`'s Ollama-not-found guidance echoed a literal `curl -fsSL https://ollama.com/install.sh | sh` string, which trips the common curl-pipe-shell SAST pattern even though it's just displayed text, never executed. Pushed `0522e14`: that branch now points at `https://ollama.com/download` instead, matching the existing non-Linux fallback. Will confirm once Codacy re-analyzes.

cyberlife-coder:
Status on the Codacy `1 high` flag: I've made two real hardening passes at it (non-loopback bind now refused by default in `27edddb`, `curl | sh` guidance text removed in `0522e14`) and it's still reporting `1 high` after both. The check API (`GET /check-runs/{id}`) returns an empty `summary`/`text` for this repo, and the Codacy dashboard itself 403s without an authenticated login I don't have access to in this session — so I can't see which specific line/rule it's pointing at.

I'm not going to keep guessing further at this without a concrete pointer — the two changes made are real improvements either way (unauthenticated-by-default network exposure and an unaudited-install-pattern in user-facing text are both legitimate to fix regardless of whether either was Codacy's exact target). One remaining plausible candidate I'm deliberately *not* acting on: `serve_http` runs plain HTTP (no in-process TLS) — but that's the documented, intentional design for a loopback-only local daemon, and bolting on TLS for that would be disproportionate scope creep this cycle. If it's that, happy to revisit given confirmation.

This check isn't in `ci-success`'s required list (confirmed via `.github/workflows/ci.yml`), so it isn't blocking. Whoever has Codacy dashboard access — a two-line pointer to the actual finding would let me close this precisely instead of guessing.

PR #1572, opened 2026-07-24
fix(hooks): pre-push exit 141 (SIGPIPE) — drain refs, exact branch match, HTTPS push doc

cyberlife-coder:
## Problem

Since #1540 made `.githooks/pre-push` executable, pushes that trigger the full validation fail with a **silent exit 141 (SIGPIPE)** right after `Pre-push validation PASSED!` — the transfer never happens. Reproduced 3× on 2026-07-23 evening; `--no-verify` works.

## Root cause (reproduced, instrumented)

`git push` opens the connection to the remote and reads the ref advertisement **before** running the pre-push hook, then sends the pack **on the same connection**. The full validation takes ~15 min (cargo test alone: 865 s warm). An instrumented repro (`GIT_TRACE_PACKET` + `ssh -vv`, 900 s delay hook) shows GitHub closing the idle SSH push session after **~360 s** (a `keepalive@openssh.com` probe, then channel EOF, `Connection to github.com closed by remote host`). When the hook finishes, git writes the push commands into the dead pipe → SIGPIPE → exit 141 with no error message.

Notes from the investigation:
- A raw idle `ssh git-receive-pack` connection survives 25 min; only real push sessions are reaped early (~6 min), so client keepalives were not relied upon as a fix.
- The old glob matched **any** ref containing `main`/`develop` — e.g. `chore/backmerge-main-*` — so ordinary chore branches also paid the 15-min validation and hit the bug.
- Checks ran once **per matching ref**, cargo children inherited git's ref pipe as stdin, and stdin was left undrained when a check failed.

## Fix

- **Hook**: drain the ref list from stdin fully before any check, match protected branches exactly (`refs/heads/main|develop`), run the validation once, give every cargo invocation `</dev/null`, and document the mechanism in the header.
- **Transport (documented in CONTRIBUTING.md / PROJECT_STRUCTURE.md)**: configure an HTTPS push URL:
  ```
  git config remote.origin.pushurl https://github.com/cyberlife-coder/VelesDB.git
  git config credential.helper '!gh auth git-credential'
  ```
  Over stateless HTTP the ref advertisement and the pack upload are independent requests, so a connection dropped during the validation window is harmless. Fetch stays on SSH.

## Verification

- A/B with a 900 s delay hook on a real push: **SSH → exit 141** at 902 s (connection died at 360 s); **HTTPS → clean protocol response** after the same delay.
- This branch itself was pushed over HTTPS **with the 900 s delay hook active and no `--no-verify`** — end-to-end proof.
- Hook logic tested with stubbed cargo: single validation run for a protected ref among several, fast path for feature branches and for names like `maintenance`, cargo stdin verified empty.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -6,6 +6,18 @@
 # Ce hook s'exécute AVANT chaque git push.
 #
 # Pour activer: git config core.hooksPath .githooks
+#
+# NOTE SIGPIPE (exit 141): git ouvre la connexion au remote AVANT de lancer ce
+# hook et envoie le pack APRÈS sur la même connexion. Si cette connexion SSH
+# meurt pendant la validation complète (~15 min), git écrit dans un pipe fermé
+# et sort en 141 sans message, après "PASSED" — le transfert n'a jamais lieu.
+# Contre-mesures ici:
+#   - stdin (liste des refs) est drainé AVANT les checks, et les sous-processus
+#     cargo reçoivent </dev/null pour ne pas consommer le pipe de git;
+#   - pousser en HTTPS rend la coupure sans effet (advertisement et envoi du
+#     pack sont deux requêtes indépendantes, la connexion est rétablie):
+#     git config remote.origin.pushurl https://github.com/cyberlife-coder/VelesDB.git
+#     git config credential.helper '!gh auth git-credential'
 # =============================================================================
 
 set -e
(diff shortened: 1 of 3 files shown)


cyberlife-coder:
Superseded by #1574, with credit: the hook body rewrite from this branch already reached develop through #1567 (committed from this session's working tree), and #1574 now carries this PR's remaining delta — the refined instrumented diagnosis (idle-timeout refuted → HTTPS pushurl as the structural countermeasure) and the CONTRIBUTING/PROJECT_STRUCTURE notes — applied on current develop to avoid rebasing a branch that is still checked out in the investigating session's worktree. Nothing from the investigation is lost.

PR #1733, opened 2026-08-01
fix(core): TRUNCATE reports actual removed edge count, not the pre-delete tally

cyberlife-coder:

`fix/*` → `develop`. Matches Git Flow.

## Description

`Database::truncate_graph` computed `deleted_edges` from `edges.len()` — the
size of the edge snapshot fetched *before* the removal loop — instead of
counting what `remove_edge` actually removed. `remove_edge` returns `false`
(no panic, fail-closed) when, for example, the WAL append for the remove
intent fails; in that case the edge stays in the store but the old code still
reported it as deleted. TRUNCATE could therefore claim a fully-cleared graph
while some edges silently remained.

Fix: count only the edges `remove_edge` reports as actually removed, via a
small extracted unit — `count_successful_removals(ids, remove)` — that takes
the removal as an injected closure and is independently testable.

This also happened to be the first execution-level test for graph TRUNCATE
via the DDL executor — the existing coverage (`admin_ddl_tests.rs`) only
exercised VelesQL parsing, not `execute_truncate`/`truncate_graph` behavior.

## Type of Change

## How Has This Been Tested?

- [x] Unit tests:
  - `test_truncate_graph_reports_actual_removed_edge_count` — TRUNCATE on a
    graph collection with nodes and edges end-to-end via
    `Database::execute_query`. Nominal-path coverage; the happy path can't
    distinguish the fix from the original bug on its own.
  - `test_count_successful_removals_counts_only_successes_without_short_circuiting` —
    drives the extracted counting logic directly with an injected closure
    that fails on every other id, and asserts both the resulting count and
    that every id was still attempted. This is the actual regression test:
    it cannot compile against `develop`, where the function doesn't exist,
    so it is red-then-green against this fix specifically.

**Test Configuration:**
- OS: Linux

Also ran locally before pushing:
- `cargo fmt --all`
- `cargo clippy -p velesdb-core --all-targets --features persistence,gpu,update-check -- -D warnings -D clippy::pedantic` (clean)
- `cargo test -p velesdb-core --features persistence -- --test-threads=1` (full suite green)
- `cargo check -p velesdb-core --no-default-features` (green)

## Checklist

## Unsafe Code Checklist

N/A — no `unsafe` added or touched.

## High-Risk Change Checklist

N/A — does not touch `hnsw`, `storage`, `Drop`, `unsafe`, or SIMD dispatch paths.

## Additional Notes

Scope is intentionally narrow: the counting fix plus its regression test.
No other behavior in `truncate_graph`/`execute_truncate` was changed. A
related issue (nodes propagate `TRUNCATE` errors via `?` while edges stay
silent, so a partial truncate can still report `Ok`) was flagged in review
and filed separately — deliberately not folded into this PR.


cyberlife-coder:

cyberlife-coder:
> **Demande de changements** — postee en commentaire : GitHub refuse une revue `REQUEST_CHANGES` sur une PR du meme compte. A traiter comme bloquant.

Le correctif est juste, le perimetre est sain, et je le prendrais volontiers. **Le test, lui, ne prouve pas ce que la PR lui fait dire** — c'est le seul point bloquant.

## Le correctif : rien a redire

Le passage de

```rust
let edge_count = edges.len();
for edge in &edges { let _ = gc.remove_edge(edge.id()); }
```

a

```rust
let edge_count = edges.iter().filter(|edge| gc.remove_edge(edge.id())).count();
```

est correct, y compris sur le point ou ce genre de reecriture derape d'habitude : `count()` consomme l'iterateur en entier, donc **chaque arete est toujours tentee**. Il n'y a pas de court-circuit — ce qui aurait ete le cas avec `.all()` ou `.take_while()`.

## Le point bloquant

La case cochee dit :

> [x] I have added tests that prove my fix is effective

`test_truncate_graph_reports_actual_removed_edge_count` cree deux aretes, tronque, et assere `deleted_edges == 2`. **Aucune defaillance de `remove_edge` n'est injectee**, donc les deux suppressions reussissent, et l'ancien code aurait rendu `2` lui aussi : `edges.len()` vaut 2.

**Ce test passe a l'identique sur `develop`, sans le correctif.** Il ne peut pas distinguer les deux versions.

Son propre commentaire le dit sans le voir :

> the two diverge whenever `remove_edge` fails on an individual edge

— precisement la condition que le test ne cree jamais.

Ce n'est pas une chicane de forme : c'est la difference entre un test de **caracterisation** et un test de **regression**. Le premier documente le comportement actuel ; seul le second empeche la regression de revenir. En l'etat, quelqu'un qui reintroduirait `edges.len()` demain verrait la suite rester verte.

La PR a une vraie valeur de couverture par ailleurs — elle le dit justement, c'est le premier test d'execution de `TRUNCATE` sur un graphe via le `DdlExecutor`, et cela merite d'etre garde.

## Ce que je demande

**Un test qui echoue sur `develop` et passe avec la modification.** Le point dur est de faire rendre `false` a `remove_edge` de facon deterministe. `ConcurrentEdgeStore::remove_edge` (`crates/velesdb-core/src/collection/graph/edge_concurrent/mod.rs:354-372`) a deux sorties `false` :

- `:359-361` — l'id est absent de la table `edge_ids` ;
- `:366-370` — l'id est dans la table mais le shard n'a plus l'arete (desynchronisation d'index), et ce chemin **repare** la table au passage.

Deux facons d'y arriver, au choix :

1. **Une couture.** Extraire le comptage dans une petite unite qui prend le retrait en parametre (closure ou trait), et la tester avec un retrait qui echoue sur une arete sur deux. Le test devient deterministe et rouge-d'abord sans toucher au stockage.
2. **Le niveau `ConcurrentEdgeStore`.** Construire l'etat que `:366-370` decrit, puis verifier que `truncate_graph` compte 1 la ou l'instantane en contient 2.

Si aucune des deux n'est atteignable a cout raisonnable, la sortie honnete est d'**enlever la case cochee** et d'ecrire dans le corps que le test couvre le chemin nominal sans prouver le correctif — plutot que de laisser une affirmation que le code ne soutient pas.

## Hors perimetre, et a garder hors perimetre

En lisant la fonction, une seconde faiblesse apparait : les nœuds propagent leur echec (`gc.delete(&node_ids)?`) alors que les aretes restent muettes, donc un `TRUNCATE` partiel rend un compte honnete **et** un `Ok`. Cela merite sa propre correction — j'ai ouvert une issue separee.

**Ne pas l'ajouter a cette PR.** Son perimetre etroit est une qualite, pas un manque.

cyberlife-coder:
Fixed in d9b2b1f, via your option 1.

Extracted the counting into `count_successful_removals(ids, remove)` — a pure function taking the removal as an injected closure. Its test forces `remove` to fail on every other id and asserts both the resulting count and that every id was still attempted (no short-circuit from `filter`+`count`). That test doesn't compile against `develop`, since the function doesn't exist there — red-then-green as requested, no fault injection into `ConcurrentEdgeStore` needed.

`test_truncate_graph_reports_actual_removed_edge_count` stays as-is, now documented as nominal-path coverage (first execution-level TRUNCATE test for graph collections) rather than proof of the fix — the two tests now have distinct, honest jobs.

The node-error-swallowing issue is out of scope here as you said; left untouched.

PR #1806, opened 2026-08-04
fix(memory): harden shared MCP session wiring

cyberlife-coder:
## Summary

- prove that an idle-expired MCP session is rejected by the daemon within one second, before any memory tool or Ollama call
- wire Codex 0.113+ directly through native Streamable HTTP so it can re-initialize after an expired-session `404`
- pin Claude Desktop to `mcp-remote@0.1.38 --transport http-only`, enforce its current Node.js >=20.18.1 runtime floor, and preserve strict CA verification
- make the shell and PowerShell installers non-destructive on failed client updates, including both PowerShell native-command error modes
- document the diagnosed lifecycle boundary and the remaining Desktop restart requirement honestly

## Root cause

The daemon expires an idle session and returns `404 Not Found` immediately, before dispatching `memory_scope` or any other tool. `mcp-remote@0.1.38` consumes that response without completing the pending JSON-RPC request or establishing a new session, so the host waits until its own timeout. This is bridge lifecycle latency, not VelesDB recall or Ollama latency.

Codex 0.113+ has native Streamable HTTP recovery and is therefore configured without the bridge. Claude Desktop still needs the stdio bridge in its local config; after an idle-session timeout it must be fully restarted.

## Verification

- `cargo test -p velesdb-memory --features http --test http_transport` — 13 passed
- targeted expired-session regression — passed, rejection bounded to one second
- installer contracts — 17 passed
- complete script suite — 321 passed
- targeted `cargo clippy -D warnings`
- Bash syntax and PowerShell parser checks
- PowerShell CRLF and `git diff --check`
- full repository pre-commit gates on the Rust commit

## Remaining limitation

Claude Desktop's pinned bridge does not transparently recover after its MCP session expires. Restart Desktop after that timeout; `http-only` prevents fallback but does not fix the bridge lifecycle bug.


cyberlife-coder:
Superseded by #1807 with the exact same head commit (`bc3e77cad8dc9474d02e1123c9c41e84b3238c01`) under the repository-compliant `fix/*` branch naming convention. No code changed.

PR #1948, opened 2026-08-16
feat(memory): constrain extraction decoding with the schema it already states [blocked]

cyberlife-coder:

`feat/*` → targets `develop`. ✅

## Status: draft, blocked on #1956

**Do not merge as it stands.** It sends `format` unconditionally, and that is measured to be wrong for at least one model.

The measurement tooling this PR originally carried has been split into **#1955**, which touches no crate code and can land on its own. Once #1955 merges, this PR's diff shrinks to the product change and its documentation. Nothing needs rebasing for that — `develop` will simply already contain the rest.

## What the evidence says

Constrained decoding is a large gain on small models and a regression on the recommended one. Measured on a CPU runner, single pass, **quality counts only** — that environment says nothing usable about latency:

| model | reference (fatal / parse) | constrained (fatal / parse) |
|---|---|---|
| `qwen3:0.6b` | 5 / 0.83 | 1 / **1.00** |
| `llama3.1:8b` | 24 / **0.00** | 2 / **0.96** |
| `qwen3:8b` | 2 / 0.96 | **0** / **1.00** |
| `qwen3:14b` | **0** / **1.00** | **1** / 0.96 |

`llama3.1:8b` going from *no valid reply at all* to 96% is the headline, and it reproduces the campaign's finding against the schema the crate would actually send.

`qwen3:14b` going the other way is the blocker. Its `truncated` count moves 0 → 1 in the same step, which is a mechanism rather than a fluctuation: the grammar requires all three sections including `facts`, replies lengthen, and a 512-token cap set against shorter replies cuts one. A cut reply does not parse.

A separate check confirmed the schema must declare `facts`: a property a schema does not name does not survive llama.cpp's grammar at all, so the campaign's variant — which omitted it — produced replies with no facts and scored models that had been prevented from answering half the question.

## What has to happen first

**#1956** — freeze `MAX_GENERATION_TOKENS` on an uncensored measurement. 512 already truncates 16 of 432 replies in the *published* campaign, before constrained decoding enters into it, and the observed maximum is the cap itself, so the value cannot be chosen from the data in hand. That needs a machine that can afford high-cap runs.

If the cap explains the whole regression, there is no size-dependent rule to encode anywhere and `format` is simply good. If it does not, constrained decoding becomes an attribute of the user-selected skill (**#1957**) rather than a library default — a setting with opposite effects depending on the model is the definition of something that cannot be a global policy.

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- [x] Unit tests
- `cargo test -p velesdb-memory --all-features` — 796 passed, 0 failed
- `cargo clippy -p velesdb-memory --all-targets --all-features -- -D warnings -D clippy::pedantic` — clean, and again with default features
- `cargo fmt -p velesdb-memory -- --check` — clean

Three tests were added and the request body moved into a pure builder, so the wire contract is assertable without a live model: a `format` that silently stopped being sent would restore the exact defect this closes, and no stub-backed test would notice.

**Test Configuration:**
- OS: Linux (container)

## Checklist

- [ ] Any dependent changes have been merged and published — **#1955 first, then #1956 decides this PR's shape**

## Unsafe Code Checklist

Skipped — no `unsafe` code touched.

## High-Risk Change Checklist

Skipped — no `hnsw`, `storage`, `Drop`, `unsafe`, or SIMD dispatch paths touched.

## Additional Notes

Constrained decoding **repairs structure, not comprehension**: a constrained small model returns valid JSON that is more often wrong about the passage. The guide and CHANGELOG both say so, because "the 8 GB tier now works" is the misreading this change makes easiest.


cyberlife-coder:
Closing this draft. Nothing here is being reversed or reconsidered — the blocker stands exactly as this PR states it, and closing is about not carrying an unmergeable draft in the open-PR list for a decision that cannot be taken yet.

**Where the two dependencies stand:**

- **#1955 is merged** (2026-08-17). The measurement tooling is on `develop`, so the diff left here is only the product change and its documentation, as this PR anticipated.
- **#1956 is still open, and cannot be resolved in CI.** It needs an uncensored high-cap run on at least one small and one large model. A 4096-token arm on a 14B is out of reach on a four-vCPU GitHub runner without a GPU — that has already been attempted and failed. This needs real hardware, not another CI attempt.

**Why this is not merged instead.** The default it ships is measured wrong for `qwen3:14b`, the recommended model: `fatal` 0 → 1 and `truncated` 0 → 1 in the same step. That is a mechanism, not noise — the schema requires `facts`, replies lengthen, and the 512-token cap cuts one. Merging would trade a regression on the recommended tier for a gain on tiers below it, on the strength of a cap value that #1956 says cannot be chosen from the data in hand.

**Nothing is lost by closing:**

- The branch `feat/memory-extractor-constrained-decoding` is untouched — reopening is one click.
- #1956 carries the blocker, the censored-distribution argument, and the success criterion.
- #1957 carries the alternative shape: if the cap does *not* explain the regression, constrained decoding belongs to the user-selected skill rather than being a library default, because a setting with opposite effects depending on the model cannot be a global policy.

Reopen when #1956 has its measurement. If the cap explains the whole regression, this PR merges close to as-is; if it does not, the change belongs in #1957 and this branch is the wrong shape for it either way.

---

PR #2004, opened 2026-08-19
refactor(core): move the remaining crate-root, metrics/, gpu/, storage/, update_check/, agent/ and simd_native/ inline test modules to sibling files

cyberlife-coder:

`refactor/*` → targets `develop`. ✅

## Description

Fourth and last lot for **velesdb-core** (#1918), after `collection/` (#2001), `index/`+`sparse_index/`+`quantization/` (#2002) and `velesql/`+`filter/`+`wire/`+`api_types/` (#2003, in flight). Twenty-five files — and the two shapes the up-front audit had flagged as the hardest are both in this lot.

**`test_fixtures.rs` holds a `pub(crate) mod fixtures`, not a private `mod tests`.** Eight test files across the crate import `crate::test_fixtures::fixtures::{setup_collection, make_point_with_payload, …}`. The extraction tool refuses this shape rather than emit a wrong declaration, so it was moved by hand with `pub(crate)` reproduced verbatim; silently dropping the visibility would have broken all eight call sites.

**`simd_neon.rs` carries its block under `#[cfg(all(test, target_arch = "aarch64"))]`,** which an x86 build never compiles — so the test-name diff, which is this campaign's main evidence, says nothing at all about it. I had expected to have to ship it on CI's word alone. Instead the aarch64 `rust-std` component turned out to be installable here, so the crate was cross-checked with `cargo check -p velesdb-core --target aarch64-unknown-linux-gnu --profile test`, which does compile the code behind that gate. It passes.

**Five files already had a `<stem>_tests.rs` sibling** (`config`, `distance`, `gpu_csr`, `half_precision`, `simd_dispatch`). `config.rs` and `distance.rs` name their modules `shared_toml_tests` and `distance_metric_names_tests`, so those names resolve the collision on their own; the other three take `_unit_tests.rs`.

With #2003 and this PR merged, velesdb-core is clear and only tauri-plugin-velesdb (5 files) remains on the frozen baseline.

## Type of Change

- [x] ♻️ Refactor (no behavior change)

## How Has This Been Tested?

- **Sorted test-name list byte-identical before/after** (`-- --list`, diff empty), re-verified after rustfmt — 5 357 tests
- `cargo test -p velesdb-core --lib` — 5 341 passed, 0 failed, 16 pre-existing ignored
- `cargo check -p velesdb-core --target aarch64-unknown-linux-gnu --profile test` — clean (covers the NEON-gated block that x86 cannot reach)
- `python3 scripts/check-inline-tests.py --root .` — PASSED (16 files / 2 229 lines)
- `python3 scripts/check-file-budgets.py --root .` — PASSED
- `cargo clippy -p velesdb-core --all-targets` — clean
- `cargo fmt -p velesdb-core --check` — clean

## Checklist

## Unsafe Code Checklist

N/A for the change itself — test-module relocation only. `simd_neon.rs`'s `unsafe` NEON intrinsics were not touched; the block that exercises them moved file unchanged, and still compiles for aarch64 (see above).

## High-Risk Change Checklist

N/A — no production code changed.


cyberlife-coder:
CI caught something the campaign's own checks could not, so recording it here.

`Lint & Format` failed on `check_prod_unwraps.py`: three `.expect()` calls in `crates/velesdb-core/src/fixtures.rs` read as production code.

The guard decides test-vs-production two ways — by filename (`*_tests.rs`, `*_test.rs`, or a path under `tests/` / `benches/`) and by skipping `#[cfg(test)]` regions inside a file. The fixtures body satisfied the second rule while it was inline in `test_fixtures.rs`. Moving it to a sibling named `fixtures.rs` took it out of *both* rules at once: no longer inside a gated region, and not named like test code. The `.expect()` calls never changed — only how the guard classified the file around them.

Fixed by naming the sibling `fixtures_tests.rs`, which restores the exemption through the repository's own convention. Test list still byte-identical at 5 357.

Worth flagging separately: `crates/velesdb-core/src/index/io_counters.rs`, added by #2002, has the same shape — test-only code under a production-looking name. It passes today only because it happens to contain no `unwrap` or `expect`; the trap is armed for whoever edits it next. I have not renamed it here to keep this PR scoped, but it is a one-line follow-up (`#[path]` in `wal_framing.rs` plus the file name) and I am happy to open it.

PR #2005, opened 2026-08-19
refactor(tauri-plugin): move the five inline test modules to sibling files

cyberlife-coder:

`refactor/*` → targets `develop`. ✅

## Description

Ninth and last crate of the #1918 campaign. All five blocks are a plain `#[cfg(test)] mod tests` at end of file, top-level, with no sibling to collide with — every one is a straight `#[path]` move.

**Verification here is weaker than for the other eight crates, and the difference is worth stating plainly rather than glossing.** This crate does not build in this environment: `cargo check -p tauri-plugin-velesdb` fails in the `gdk-sys` build script for want of the GTK development libraries. So there is no test-name list to diff — the campaign's primary evidence everywhere else — no `cargo test`, and no clippy.

What did run, and what it is worth:

- `cargo fmt --check` is clean, but rustfmt only parses; it proves the files are syntactically valid Rust, not that the modules resolve.
- The three text-level guards pass (`check-inline-tests.py`, `check-file-budgets.py`, `check_prod_unwraps.py`). The last one is checked here deliberately: in #2004 a moved test module fell out of that guard's exemptions and turned a green refactor red, so it is now part of the routine.

The compile-level proof rests on CI. If CI disagrees with any of the above, the failure is real and mine.

**With this merged, the inline-tests baseline reaches zero** — 173 files and 21 788 frozen lines when #1918 opened, none left. Note this PR's guard count reads 25 files rather than 5 because it branches from `develop` before #2004 (core lot D) lands; #2004 removes the other 20.

## Type of Change

- [x] ♻️ Refactor (no behavior change)

## How Has This Been Tested?

- `python3 scripts/check-inline-tests.py --root .` — PASSED
- `python3 scripts/check-file-budgets.py --root .` — PASSED
- `python3 scripts/check_prod_unwraps.py` — PASSED
- `cargo fmt -p tauri-plugin-velesdb --check` — clean
- `cargo test` / `cargo clippy` — **not run: the crate does not build in this environment (missing GTK dev libraries).** Covered by CI only.

## Checklist

- [ ] New and existing unit tests pass locally with my changes — **cannot be checked locally, see above**

## Unsafe Code Checklist

N/A — test-module relocation only.

## High-Risk Change Checklist

N/A — no production code changed.


cyberlife-coder:
Two real CI failures on `1a62b47`, both the same root cause, now fixed.

`Lint & Format` and `Gate Contracts / Repository guard contracts` both failed on:

```
FAIL: test_the_baseline_file_is_well_formed
AssertionError: {} is not true : the checked-in baseline must not be empty
```

`TestRealBaseline` asserted the checked-in baseline was non-empty. That was true for as long as the baseline was a debt ledger — and this PR is the one that empties it. The campaign's finish line was, quite literally, an assertion failure.

Worth being precise about why dropping the assertion is safe rather than convenient: **the baseline is an exemption list, so empty is its strictest state, not its weakest.** Each entry is permission for one file to keep its inline test module. With zero entries, any inline test module anywhere in the workspace counts as new debt and the guard refuses it. A truncated baseline could never silently weaken the guard — it would fail loudly on every file it used to exempt. There was nothing for the non-emptiness check to protect.

So I removed that one assertion, left the per-entry shape checks (positive line count, no `_tests.rs` path) in place for whatever entries a future change adds, and recorded the reasoning in a comment on the test.

Verified locally: `test_check_inline_tests` 35 tests OK, plus `test_check_file_budgets`, `test_check_prod_unwraps`, `test_check_feature_claims`, `test_check_perf_claims` all OK. I could not run the full `unittest discover` locally — `test_guard_refusal_vectors` exceeds my time budget in this environment — so the complete suite is CI's word, as is everything compile-level for this crate.

One note on reading the earlier log: the `Gate Contracts` tail shows a `Performance claims audit FAILED` block just before the error. That is a red herring — it is the expected stdout of `test_two_documents_disagreeing_are_refused`, a passing test that feeds the auditor two contradictory fixture documents on purpose. The 38.60 µs vs 120.00 µs figures are fixtures, not claims in our docs.

cyberlife-coder:
Correction to the evidence I gave in my previous comment, in the PR's favour.

I said the full `unittest discover` could not run here because `test_guard_refusal_vectors` exceeds my time budget. It did in fact complete — I had started it in the background before the fix and misread its result, because I had piped it through `tail`, so the exit code I saw belonged to `tail`, not to Python.

The real result of that pre-fix run:

```
Ran 625 tests in 723.182s
FAILED (failures=1, skipped=9)
AssertionError: {} is not true : the checked-in baseline must not be empty
```

So the whole guard contract suite — all 625 tests — was exercised against the empty baseline, and `test_the_baseline_file_is_well_formed` was the *only* failure. That is a stronger statement than the per-module runs I listed before: nothing else in the suite depends on the baseline being non-empty, which is what I wanted to know and had only checked by grep.

The fix itself is verified by the post-fix run of `test_check_inline_tests` (35 tests OK). I have not re-run the full 12-minute discovery after the fix, since it had no other failure to fix.

Unchanged from before: compile-level verification of this crate is still CI's alone, for the GTK reason in the PR description.

PR #2146, opened 2026-08-27
fix(metrics): normalize MAP by the corpus-wide relevant count

### LINKED ISSUE #2106: Metric-calculation audit 2026-08-22: minor findings and test-coverage gaps (batchable)
Companion to the critical/major issues #2100–#2105 from the end-to-end metric-calculation audit (distance kernels scalar/AVX2/AVX-512/NEON, quantization, fusion, observability metrics; bindings excluded). Each item below is verified with file:line evidence but is minor on its own — batchable into a few cleanup PRs.

## Score semantics

1. **Cosine score range inconsistent across paths** — HNSW `transform_score` clamps to `[0,1]` (`index/hnsw/native/backend_adapter.rs:379`) while every exact path returns `[-1,1]` (`simd_native/scalar.rs:125`, `index/hnsw/index/search.rs:63`, GPU clamp `index/rerank.rs:171`). Anti-correlated vectors collapse to 0.0 in one mode and score negative in another; delta-merge mixes the two ranges. Same family as #2095 item A2 (core vs WASM) but also intra-core.
2. **Dimension-mismatch fallback score 0.0 is a perfect score for distance metrics** — `compute_metric_score` returns 0.0 on length mismatch (`collection/search/query/extraction.rs:261-262`), so in `WHERE similarity(...) &gt; t` filtering (`similarity_filter.rs:99-100, 212-216`) a malformed named vector passes as a perfect match. The ORDER BY path already uses a metric-aware worst sentinel (`ordering.rs:239-243`) — apply the same here (exclude, or worst sentinel).
3. **`CpuDistance` reference impl diverges from shipped semantics** (test-only reach, `index/hnsw/native/distance.rs:232-259`): Hamming counts differing bit patterns instead of the 0.5-threshold semantics (and counts `0.0` vs `-0.0` as different); Jaccard returns distance 1.0 for two all-zero vectors where production yields 0.0. As the declared baseline for SIMD-correctness testing it validates the wrong semantics.
4. **NEON Jaccard NaN divergence** — `vminq/vmaxq_f32` propagate NaN (`simd_native/neon.rs:790-791, 858-877`) while scalar `f32::min/max` ignore it (`scalar.rs:153-159`): same NaN-bearing input gives finite score at dim 3 and NaN at dim ≥4 on aarch64.
5. **Public `simd_neon::cosine_neon_safe` unclamped** (`simd_neon.rs:140-150`) — can return &gt;1.0 and diverges from `cosine_similarity_native`; legacy module, internally unused.

## Unwired-but-public score_fusion modules (fix or retire)

6. `ScoreFusionMethod::Rrf` pseudo-RRF `1/(60+(1−s)·100)` divides by zero at s=1.6 and goes negative beyond (`collection/search/query/score_fusion/mod.rs:246`); the module is confirmed unwired (`tests/bdd/fusion_weighted_bug.rs:89` documents it, incl. the known `Weighted==Average` defect).
7. `BoostCombination::Max` folds from identity 1.0, ignoring all-penalty boost sets (`score_fusion/boost.rs:226`; `Min` on :227 is correct).
8. `PathScorer::score_rel_types` decays quadratically (`decay^(n(n+1)/2)`) contradicting `score_length` and its own doc (`score_fusion/path.rs:88-95`).

## SIMD hygiene

9. **Latent UB: out-of-bounds pointer probes** `p.add(N) &lt;= end_ptr` computed before knowing N elements remain — violates `pointer::add`'s contract (Miri-flaggable) though never dereferenced. Representative: `x86_avx2_similarity.rs:53, 267, 401`; `x86_avx512.rs:127, 193, 249, 355, 538, 805, 1289, 1330`. The index-based guards (`i + 16 &lt;= len`, e.g. `x86_avx512.rs:850, 948`, `x86_avx2/dot.rs`) are the correct pattern to copy.
10. **`SimdLevel::Avx512` granted on `avx512f` alone** (`simd_native/dispatch/mod.rs:56-57`) while Avx512-level arms call `#[target_feature(enable = "avx2")]` kernels without an AVX2 check (`dispatch/hamming.rs:61, 99, 133, 176, 231`; `dispatch/euclidean.rs:105`). A hypervisor can mask AVX2 with AVX512F exposed. One extra `is_x86_feature_detected!("avx2")` closes it.
11. **Sparse `linear_scan_dense` duplicate emission on exact zero cancellation** — accumulator value used as first-touch sentinel (`sparse_index/search/strategy.rs:121-124`); with negative query weights, +1.0 then −1.0 then a third posting pushes the doc into `touched` twice → duplicated result, legitimate k-th doc evicted. Use a seen-set or sentinel outside the score domain.

## Observability nits

12. **MAP normalizes by retrieved-relevant count** while its doc states textbook `AP = (1/R)·Σ` (`metrics/retrieval.rs:271-302`, division :294): retrieving 1 of 10 relevant at rank 1 scores AP=1.0. Doc/behavior mismatch — fix the doc or the signature.
13. **Millisecond histogram buckets use exclusive bounds** (`ms &lt; bound`) but export Prometheus `le` labels, which are inclusive (`collection/graph/metrics/mod.rs:72-76`, `collection/search/query/match_metrics.rs:99-104`): an exactly-1ms observation lands in the wrong bucket.

## Test-coverage gaps (would have caught the majors)

14. **Cosine/L2/dot tail paths never exercised**: `cosine_fused_tests.rs:23` uses only multiples of 16; AVX-512 4-acc/8-acc masked tails and `cosine_8acc_remainder` are dead under the suite (need dims like 257, 513, 1025, 1041); the only odd-dim cosine test uses identical vectors, for which tail-skipping is undetectable by construction. L2 4-acc/8-acc tails likewise (tests stop at len 65 / multiples of 32); dot 8-acc masked tail needs e.g. 1025. `hamming_jaccard_tests.rs:65, 169` (dims 1…1025) is the model to copy.
15. **Cosine tolerance too loose**: absolute `5e-3` on a value bounded in [-1,1] (`cosine_fused_tests.rs:14`) — at dim 768 a fully skipped tail element shifts cosine by ~1.3e-3, inside tolerance.
16. **No test covers sub-1e-3-magnitude cosine inputs** (would catch #2103) **or NaN-bearing Jaccard/Hamming inputs** (would catch item 4).
17. **`harley_seal_tests.rs` names an algorithm that does not exist** — no CSA/Harley-Seal network anywhere in the tree; binary popcount is `vcntq_u8`/VPOPCNTDQ/`count_ones` (all correct). Rename the tests to what they actually exercise (threshold-based f32 Hamming dispatch) or implement the named algorithm.
18. **NEON kernels only tested on aarch64 runners** — all equivalence tests go through `*_native` dispatch, which never executes `neon.rs` on x86 CI.
19. **Hamming/Jaccard PQ rescoring under OPQ rotation** computes the metric in rotated space (`collection/search/vector.rs:151-168`); orthogonal rotation preserves L2/dot/cosine but not Hamming/Jaccard. Exotic configuration — flagged for a guard or a documented divergence entry.

Full per-area "checked and correct" coverage lists are in the audit records for #2100–#2105; the positive results (BM25 IDF, RRF/WeightedRRF, NDCG/MRR/recall formulas, ADC/PQ/SQ8 LUT arithmetic, operational counters, threshold inversion semantics, NaN-safe sorting) held up under hand-computation.

cyberlife-coder:
## Description

Closes #2106 item 12. `mean_average_precision` took `&[Vec<bool>]` and divided by the number of relevant items **retrieved**, while its own doc comment stated the textbook formula:

```rust
/// `AP = (1/R) * Σ P(k) * rel(k)` where R is total relevant items
```

The two disagree exactly where the metric earns its keep:

| retrieved | corpus has | old AP | textbook AP |
|---|---|---|---|
| `[true]` at rank 1 | 10 relevant | **1.0** | **0.1** |

A flawless score for **10% recall**. A ranking-quality metric that cannot see what it missed will always flatter a system that returns one confident result and stops.

The issue offered "fix the doc or the signature". **The doc was not the thing to fix** — a MAP that cannot penalise missed relevant documents is not MAP, and this is the retrieval-metrics module, whose whole job is honest evaluation. Silently redefining a standard metric is the worst of the three options.

## Type of Change

- [x] 💥 Breaking change — the signature moves from `&[Vec<bool>]` to `&[(&[bool], usize)]`

## How Has This Been Tested?

- [x] Unit tests

`&[bool]` over retrieved positions simply **cannot express `R`**, so the shape had to move:

```rust
pub fn mean_average_precision(queries: &[(&[bool], usize)]) -> f64
```

Each query pairs its relevance flags with the corpus-wide relevant count. A caller with no such count reproduces the old behaviour by passing the retrieved count — **explicitly, at the call site**, rather than silently inside the metric.

A `total_relevant` below the retrieved count is a caller error: it describes a corpus holding fewer relevant items than the results contain. The denominator is raised to the retrieved count, so a bad input cannot yield `AP > 1.0` and quietly corrupt an average over many queries.

### The old tests were blind to this by construction

This is the part worth pausing on. All four pre-existing tests used `total_relevant == retrieved` — **the single case where both denominators agree.** That is why the defect survived: the suite could not have caught it.

Two new tests encode the defect itself:

| test | holds |
|---|---|
| `test_map_penalises_missed_relevant_documents` | 1-of-10 at rank 1 scores `0.1`, not `1.0` |
| `test_map_falls_as_recall_falls_with_the_ranking_held_fixed` | same ranking, same precision at every position, only corpus size varies → AP must fall |

Plus two guarding the edges: an understated `total_relevant` stays within `[0, 1]`, and a corpus with nothing relevant scores 0.

### Mutation-checked

Restoring the old normalisation (`denominator = retrieved_relevant`):

```
test test_map_falls_as_recall_falls_with_the_ranking_held_fixed ... FAILED
test test_map_penalises_missed_relevant_documents             ... FAILED
test test_map_empty                                           ... ok
test test_map_mixed                                           ... ok
test test_map_no_relevant                                     ... ok
test test_map_perfect                                         ... ok
```

The two new tests go red; **all four old ones stay green** — the clearest possible statement of why this needed a signature change and not a doc edit. `retrieval.rs` was restored from backup afterwards.

**Test Configuration:**
- OS: Linux x86_64
- Rust: 1.90

Full validation: `cargo fmt --check` clean; `cargo clippy -p velesdb-core -p velesdb-server --all-targets --features persistence,gpu,update-check -- -D warnings -D clippy::pedantic` clean (the CI's own flags); **5470 lib tests pass**; `cargo check --benches --tests --features persistence,internal-bench` clean (the gate #2144 just added); all three repo guards pass.

## Checklist

## Additional Notes

**On the breaking change.** `mean_average_precision` is public API (re-exported from `lib.rs` and `metrics/mod.rs`), but its only callers in the tree are its own tests — no server, binding, SDK or integration uses it. The migration is mechanical: `&[flags]` becomes `&[(&flags, total_relevant)]`, and passing the retrieved-relevant count as `total_relevant` gives byte-identical results to today.

The doc comment now carries the *why* rather than just the formula, including the 1-of-10 example, so the next reader does not have to rediscover which denominator was intended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **`mean_average_precision` now takes the corpus-wide relevant count, and MAP
+  penalises what a search missed (#2106 item 12).** The signature was
+  `&[Vec<bool>]` and the implementation divided by the number of relevant items
+  *retrieved*, while the doc comment stated the textbook
+  `AP = (1/R)·Sum P(k)·rel(k)` with `R` the total relevant in the corpus. The two
+  disagree exactly where the metric earns its keep: retrieving **1 of 10**
+  relevant documents, at rank 1, scored a flawless `AP = 1.0` for 10% recall. A
+  ranking-quality metric blind to what it missed always flatters a system that
+  returns one confident result and stops.
+
+  The doc was not the thing to fix. `&[bool]` over retrieved positions simply
+  cannot express `R`, so the signature moved instead: `&[(&[bool], usize)]`, each
+  query pairing its relevance flags with the corpus total. A caller with no
+  corpus-wide count reproduces the old behaviour by passing the retrieved count —
+  explicitly, at the call site, rather than silently inside the metric. A
+  `total_relevant` below the retrieved count is a caller error describing an
+  impossible corpus; the denominator is raised to the retrieved count so a bad
+  input cannot push `AP` above 1.0 and corrupt an average over many queries.
+
+  The four pre-existing tests could not have caught this: every one used
+  `total_relevant == retrieved`, the single case where both denominators agree.
+  Two new tests encode the defect itself — 1-of-10 scoring 0.1 rather than 1.0,
+  and AP falling monotonically as recall falls with the ranking held fixed — and
+  both go red when the old normalisation is restored, while all four old tests
+  stay green.
+
+
 - **The `RaBitQ` traversal backend became a codec: one quantized-precision
   state machine, two codecs.** `RaBitQPrecisionHnsw`'s concurrent
   insert/train/install machinery and its graph traversal are now the
--- a/crates/velesdb-core/src/metrics/retrieval.rs
+++ b/crates/velesdb-core/src/metrics/retrieval.rs
@@ -251,53 +251,82 @@ pub fn hit_rate<T: Eq + Hash + Copy>(query_results: &[(Vec<T>, Vec<T>)], k: usiz
 
 /// Calculates Mean Average Precision (MAP).
 ///
-/// MAP is the mean of Average Precision (AP) over all queries.
-/// AP rewards systems that return relevant items early in the result list.
+/// MAP is the mean of Average Precision (AP) over all queries. AP rewards a
+/// system that returns relevant items early *and* penalises one that misses
+/// relevant items entirely — the second half is why `total_relevant` is a
+/// parameter rather than something derived from the flags.
 ///
 /// # Formula
 ///
-/// `AP = (1/R) * Σ P(k) * rel(k)` where R is total relevant items
-/// `MAP = (1/Q) * Σ AP_q` where Q is number of queries
+/// `AP = (1/R) * Σ P(k) * rel(k)` where `R` is the total number of relevant
+/// items **in the corpus**, not the number this query happened to retrieve.
+/// `MAP = (1/Q) * Σ AP_q` where `Q` is the number of queries.
+///
+/// # Why the signature carries `total_relevant`
+///
+/// This used to take `&[Vec<bool>]` and divide by the count of relevant items
+/// *retrieved*, while its documentation stated the formula above. The two
+/// disagree exactly where the metric earns its keep: retrieving 1 of 10
+/// relevant documents, at rank 1, scored `AP = 1.0` — a perfect score for 10%
+/// recall. A ranking quality metric that cannot see what it missed will always
+/// flatter a system that returns one confident result and stops.
+///
+/// The old shape could not express the fix: `&[bool]` over retrieved positions
+/// simply does not know `R`. So the signature moved rather than the doc. Where
+/// a caller genuinely has no corpus-wide count, passing the retrieved-relevant
+/// count reproduces the previous behaviour — explicitly, at the call site,
+/// instead of silently inside the metric.
 ///
 /// # Arguments
 ///
-/// * `relevance_lists` - For each query, a list of booleans indicating relevance
-///   at each position (true = relevant, false = not relevant)
+/// * `queries` - one entry per query: the relevance flags at each retrieved
+///   position (`true` = relevant), paired with the total number of relevant
+///   items in the corpus for that query.
 ///
 /// # Returns
 ///
-/// A value between 0.0 and 1.0, where 1.0 means perfect precision at every position.
+/// A value between 0.0 and 1.0. A query whose `total_relevant` is 0 contributes
+/// 0.0: there was nothing to find, so no ranking of it can be credited.
+///
+/// `total_relevant` below the number actually retrieved is a caller error — it
+/// describes a corpus with fewer relevant items than the results contain. Rather
+/// than return an `AP > 1.0` that would quietly corrupt an average, the
+/// denominator is raised to the retrieved count, which keeps the result inside
+/// its documented range.
 #[must_use]
-pub fn mean_average_precision(relevance_lists: &[Vec<bool>]) -> f64 {
-    if relevance_lists.is_empty() {
+pub fn mean_average_precision(queries: &[(&[bool], usize)]) -> f64 {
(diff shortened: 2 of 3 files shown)


cyberlife-coder:
**`Python Integrations` failed on `77478a6f`, and it is not this PR's.**

The job died in its *Install stable Rust toolchain* step, before any test body ran. `rustup_init` panicked, and `pip` then had no compiler to build `velesdb-python` with:

```
Cargo, the Rust package manager, is not installed or is not on PATH.
This package requires Rust and Cargo to compile extensions.
   ...
error: metadata-generation-failed
╰─> from file:///home/runner/work/VelesDB/VelesDB/crates/velesdb-python
```

This PR changes one Rust function signature in `velesdb-core` (`mean_average_precision`). Nothing in that diff can affect whether `rustup` installs on a runner — the causal link is absent, not merely unlikely. `velesdb-python` has no caller of the function either; its only callers in the tree are `metrics_tests.rs`.

Re-running the failed job once, which is the one case a re-run is warranted for: a job that died at checkout/install rather than in a test. If it fails a second time the failure is real and I will treat it as this PR's.

No fix ported, because there is nothing to port — no change to this branch would alter the outcome.

PR #2163, opened 2026-08-30
perf(bm25): avoid cloning terms already present in the maps

cyberlife-coder:
## ⚠️ Target branch (Git Flow)

`perf/*` → targets `develop`. ✅

## Description

Part of the interning/wiring audit tracked in #2090 ("intern terms — the corpus vocabulary is materialized once per document"). That issue's full `TermId(u32)` dictionary is a persistence-format change requiring its own PR series; this PR ships the one step the issue calls out as "shippable independently": the `get`-before-`entry` probe, mirroring the fix already applied to the CSR label interner (#2084).

`Bm25Index::add_document` builds its per-document `term_freqs` map and updates the global `inverted_index` using `map.entry(key.clone())`. `entry()` demands an owned key, so every token paid a `String` allocation to probe the map even when the key was already present:

- `term_freqs.entry(token.clone())` — clones on every repeated word within the same document (e.g. "the", "a").
- `inverted_index.entry(term.clone())` — clones once per document per term, including for terms already in the corpus vocabulary, which is the common case once the index holds more than a handful of documents.

Both sites now do a borrowed `&str` lookup (`get_mut`) first and only clone into an owned key the first time that key is seen. No change to BM25 scoring math, stored data, or the wire/persistence format — allocation is removed only on the already-present path.

Fixes # (none — partial progress on #2090; the `TermId(u32)` dictionary and its persistence-format change are left for a separate PR)

## Type of Change

- [x] ⚡ Performance improvement

## How Has This Been Tested?

- [x] Unit tests
- [x] Manual testing

Ran the existing BM25 suite (58 tests covering tokenization, scoring, persistence roundtrip, WAL replay, concurrent inserts) unchanged — all green, proving the change is behavior-preserving:

```
cargo test -p velesdb-core --lib bm25 -- --test-threads=1
test result: ok. 58 passed; 0 failed; 2 ignored
```

Also ran, at CI-exact flags:

```
cargo fmt --all -- --check
cargo clippy -p velesdb-core --all-targets --features persistence,gpu,update-check -- -D warnings -D clippy::pedantic
python3 scripts/check_prod_unwraps.py   # PASSED
```

**Test Configuration:**
- OS: Linux
- Rust version: 1.90 (workspace MSRV)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
[...]

Issue #2090, opened 2026-08-21
perf(bm25): intern terms — the corpus vocabulary is materialized once per document


cyberlife-coder:
From the final interning/wiring audit.

**Today** (`collection/graph/bm25.rs`): `Document.term_freqs: FxHashMap<String, u32>` (:65) owns a `String` for each distinct term of that document, while `inverted_index: FxHashMap<String, PostingList>` (:108) owns the same term globally — for N documents the corpus vocabulary is materialized N times over. The ingest path makes it worse: `tokenize()` allocates a `String` per token (:151-157), `term_freqs.entry(token.clone())` clones it again (:175), and `inv_idx.entry(term.clone())` clones a third time (:204) — including for terms already present, because `entry()` demands an owned key (the same defect #2084 fixed in the CSR builder; a `get`-before-`entry` probe is the cheap first step here too, shippable independently).

**Proposed:** a `TermId(u32)` dictionary — the crate already has exactly this shape in `column_store::StringTable` (Arc-shared since #2081). `Document.term_freqs` becomes `FxHashMap<TermId, u32>`, `inverted_index` keys become `TermId`, query terms resolve once per query.

**Constraint that makes this an issue and not a patch:** BM25 documents and the inverted index participate in persistence — the dictionary must be persisted with them (or rebuilt at load), and the WAL/snapshot format versioned accordingly. The sparse index next door (`sparse_index/inverted_index.rs`) already uses u32 term ids and is the template.

**Validation bar:** BM25 scoring bit-identical on a fixture corpus (IDF/TF math unchanged — only key representation moves); persistence roundtrip suites; RSS before/after on a corpus where vocabulary ≪ total tokens; the `get`-before-`entry` step alone is measurable by allocation count.

cyberlife-coder:
Delivered by #2181 (squash-merged to `develop` as 303fac6a), following the measure-first discipline: baseline posted above, then the interning, then the same harness re-run.

Final numbers (20k docs × 120 tokens, 5k-term vocabulary, release, counting-`GlobalAlloc`):

| Metric | Before | After | Δ |
|---|---|---|---|
| Live memory after ingest | 215.1 MB (10,756 B/doc) | 74.1 MB (3,704 B/doc) | **−65.6%** |
| Allocations per doc | 255.1 | 136.5 | −46% |
| Ingest (20k docs) | 1.35 s | 0.91 s | −33% |
[...]

Issue #2089, opened 2026-08-21
perf(graph): wire the LabelTable — edge labels are stored 4x per edge and allocated per traversal hop


cyberlife-coder:
From the final interning/wiring audit. Three findings that share one root cause and one fix, plus the fix already exists as dead code:

**The disease:**
- `edge.rs:258-274` — `insert_edge` stores the same label string 4× per edge: `edge.label().to_string()` for the outgoing composite key, `label.clone()` for `entry()`, the `GraphEdge`'s own copy, and a second `to_string()` for `incoming_by_label`. At 10M edges from a ~20-label vocabulary this is the ~200 MB the `LabelTable` doc comment promises to save, still being paid.
- `edge.rs:383,407` — `get_outgoing_by_label`/`get_incoming_by_label` build `&(node_id, label.to_string())` **per lookup**: the maps are keyed `(u64, String)`, and no `Borrow` path exists into a tuple, so every `-[:TYPE]->` / `<-[:TYPE]-` hop pays a malloc+memcpy+free (called from `edge_concurrent/query.rs:125,137`).
- `edge_removal.rs:92,96,118` — three more `label.to_string()` per edge removal, two of them building the identical key twice; cascading node deletion multiplies by degree.

**The unwired fix:** `collection/graph/label_table.rs` is a complete interning subsystem with **zero production callers**. `ConcurrentEdgeStore` holds it behind a `RwLock` (`edge_concurrent/mod.rs:106`), constructs it empty, takes a read lock on it in the snapshot path (with a lock-ordering comment defending a lock nobody contends), and passes it to `SnapshotBuilder::build(_label_table: &LabelTable)` — **underscore-prefixed and discarded** (the builder does its own local interning; its per-edge allocation was fixed in #2084). `edge_persistence.rs:28` constructs a throwaway empty one just to satisfy the signature. This is load-bearing-looking inert machinery — the same shape as the removed PDX block.

[...]

Issue #2075, opened 2026-08-21
perf(hnsw): flat CSR adjacency for layer 0 — retire the per-node RwLock<Vec<NodeId>>


cyberlife-coder:
> **Status 2026-09-05 — one of the sub-items is settled negatively, by measurement.**
>
> Shipped: the instrument only — #2164 (`98dc21a0`, fixed by `f12b3e3d` in #2165), `benches/hnsw_adjacency_scale.rs`.
>
> **Do not relitigate the `NodeId` narrowing.** `usize` → `u32` measured +2.6 % at 200K and indistinguishable at 1M; the branch was never pushed. The sentence below about `NodeId = usize` doubling the in-memory footprint is a footprint statement, not a latency gain — it was tested and did not pay.
>
> Still true and untouched: `index/hnsw/native/layer.rs:8` (`pub type NodeId = usize;`) and `:14` (`neighbors: Vec<RwLock<Vec<NodeId>>>`), read under lock via `with_neighbors` (`layer.rs:44`) on the hot path at `graph/search.rs:404`, `:686`, `graph/search_pipeline.rs:95`, `:130` and `quantized_traversal.rs:94`. Neither the flattening nor the lock removal has been attempted, and the null narrowing result says nothing about either.

---

From the four-domain perf audit (Tier B), deliberately deferred from the perf waves because it needs wall-clock benches and a careful mutation-path design, not just the eval-count harness.

**Today**: `Layer { neighbors: Vec<RwLock<Vec<NodeId>>> }` costs one parking_lot read acquire/release (two atomic RMWs on a shared line) **plus** a pointer chase into a separately heap-allocated `Vec` for every node visit in the search loop (`with_neighbors` at `graph/search.rs`, `search_pipeline.rs`). At 1M nodes that is 1M locks + 1M scattered allocations, zero adjacency locality. `NodeId = usize` also doubles the in-memory footprint vs the on-disk `u32` format.

**Proposed**: flat CSR (`offsets: Vec<u32>` + `neighbors: Vec<u32>`) for layer-0 reads behind an epoch/`arc_swap` snapshot (the design the edge store's `CsrSnapshot` already uses), with the write path mutating a build-side structure and publishing. Search then reads a contiguous `&[u32]` slice with no per-node atomics.

[...]

PR #2175, opened 2026-08-31
perf(graph): stop allocating a label String per traversal hop (#2089)

cyberlife-coder:

`perf/2089-label-lookup-allocations` → **`develop`**. ✅

## Description

Partial fix for #2089 — the half that carries no persistence-format risk.

`incoming_by_label` was keyed `(u64, String)`. No `Borrow` path exists into a tuple, so every `<-[:TYPE]-` traversal hop allocated a `String` purely to build a probe key, and `purge_incoming_index` allocated **two** per call, one to look up and one to remove. Nesting the index as `target -> label -> Vec<edge_id>` lets `HashMap<String, _>::get` borrow the `&str` directly: those lookups and the purge now allocate nothing.

`by_label` needed no type change. `entry()` demands an owned key, so it allocated even for a label the map already held; probing with `get_mut` first removes that clone on every insert after warmup, the label vocabulary being tiny next to the edge count. Same shape as #2084 and #2163.

Net effect per edge insert: **two of the four label copies are gone**. Per `<-[:TYPE]-` hop and per edge removal: **all probe-key allocations are gone**.

### What this deliberately does not do, and why

#2089 is named for wiring the `LabelTable`, and that is not here. While implementing, I found a constraint the issue does not record and posted it there in full ([comment](https://github.com/cyberlife-coder/VelesDB/issues/2089#issuecomment-5479774236)):

`by_label` and `outgoing_by_label` are **serialized** fields of `EdgeStore`. Persistence goes through `PostcardPersistence`, a blanket trait over raw postcard bytes — postcard is not self-describing and **there is no format version anywhere in the path**. The existing `#[serde(skip)]` comment on `incoming_by_label` already says as much. Worse, `helpers.rs` records that a file `load_from_file` rejects means *"callers would fall back to an empty store, losing data"* — so a key-type change on a serialized index does not fail loudly, it can present as an empty graph.

So this PR touches only what is provably format-neutral: `incoming_by_label` is `#[serde(skip)]` and rebuilt on load, and `by_label` keeps its type. `outgoing_by_label` and the `LabelTable` interning proper stay open on #2089, pending the same format decision #2090 needs for BM25.

Partially addresses #2089.

## Type of Change

- [x] ⚡ Performance improvement

## How Has This Been Tested?

- [x] Unit tests

`cargo test -p velesdb-core --features persistence -- --test-threads=1` — **6473 passed, 0 failed** across 69 suites (lib suite 5436, up from 5434 by the two tests added here).

`cargo clippy -p velesdb-core --all-targets --features persistence,gpu,update-check -- -D warnings -D clippy::pedantic` — clean. `cargo fmt --all` applied.

Two tests added, both covering behaviour this change is responsible for:

- `test_incoming_label_index_prunes_emptied_entries` — the nested index must drop a label once its last edge goes, and drop the node entry once its last label goes. The composite key never pruned the node level at all, so this is new behaviour, not a regression guard.
- `test_rebuild_incoming_label_index_matches_live_edges` — rebuilding from live edges reproduces what the inserts built. This is the path a postcard load actually takes, since the index is `serde(skip)`.

**Test Configuration:**
- OS: Linux x86_64
- Rust version: 1.90

## Checklist

## Unsafe Code Checklist

No `unsafe` added or modified.

## High-Risk Change Checklist

Not a `hnsw`, `storage`, `Drop`, `unsafe` or SIMD path, but it does touch an index that participates in persistence, so:

- [x] **Invariants impacted.** `incoming_by_label` must hold exactly the edge ids whose target is the outer key and whose label is the inner key; it must contain no empty `Vec`, no empty inner map, and no node entry without labels. The last two are newly enforced and covered by the added test.
- [x] **Durability semantics.** Unchanged, and that is the point: no serialized field moves. `incoming_by_label` remains `#[serde(skip)]` and derived, so `edge_store.bin` keeps the layout `develop` writes and reads.
- [x] **Lifecycle tests.** The rebuild path — what runs after a snapshot load — is covered by the second added test.
- [ ] Two reviewers: leaving to maintainer judgement, given no serialized field changes.

## Additional Notes

No wall-clock figure is claimed here on purpose. The change removes allocations, which is provable by inspection; whether that shows up as latency at a given graph size is a separate question, and this container had just spent 450 s compiling and running the suite. Measuring on a machine in that state is precisely the error documented at length on #2075, where a busy container showed 22–25 % spread against an idle floor of 0.35 %. If a number is wanted, `edge_store_benchmark` (`add_edge`, `cascade_delete`) is the right instrument, run on an idle machine with each arm run twice.

### DIFF
--- a/crates/velesdb-core/src/collection/graph/edge.rs
+++ b/crates/velesdb-core/src/collection/graph/edge.rs
@@ -133,9 +133,30 @@ pub struct EdgeStore {
     pub(super) outgoing: HashMap<u64, Vec<u64>>,
     /// Incoming edges: target_id -> Vec<edge_id>
     pub(super) incoming: HashMap<u64, Vec<u64>>,
-    /// Secondary index: label -> Vec<edge_id> for fast label queries
+    /// Secondary index: label -> Vec<edge_id> for fast label queries.
+    ///
+    /// **Serialized — changing this type is a format break.** See the note on
+    /// `outgoing_by_label` below; it applies to both.
     pub(super) by_label: HashMap<String, Vec<u64>>,
-    /// Composite index: (source_id, label) -> Vec<edge_id> for fast filtered traversal
+    /// Composite index: (source_id, label) -> Vec<edge_id> for fast filtered
+    /// traversal.
+    ///
+    /// **Serialized — changing this type is a format break, and a silent one.**
+    /// This field and `by_label` are written into `edge_store.bin` by postcard,
+    /// which is not self-describing, and nothing in the path carries a format
+    /// version: `PostcardPersistence` is a blanket trait over raw bytes. A file
+    /// the new code cannot deserialize does not surface as a loud error either
+    /// — per `helpers.rs`, callers fall back to an **empty store**, so the
+    /// symptom is a graph that lost its edges.
+    ///
+    /// This is why #2089 stopped short here. Keying these by `LabelId` (the
+    /// unwired `LabelTable`) or nesting them the way `incoming_by_label` is
+    /// nested below would remove the last per-hop `String` allocation on
+    /// `-[:TYPE]->`, and it is worth doing — but only behind a format decision,
+    /// the same one #2090 needs for the BM25 dictionary. Marking them
+    /// `#[serde(skip)]` and rebuilding on load is a defensible answer, and is
+    /// still a format change, since removing a field moves the postcard layout
+    /// exactly as changing one does.
     pub(super) outgoing_by_label: HashMap<(u64, String), Vec<u64>>,
     /// Nested index: target_id -> label -> Vec<edge_id> — the incoming
     /// mirror of `outgoing_by_label`, so `<-[:TYPE]-` patterns stop paying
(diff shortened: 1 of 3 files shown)


cyberlife-coder:
Superseded by #2171, now merged into `develop` (d775ebd). That change carries this PR's wins and the half it deliberately deferred: `incoming_by_label` AND `outgoing_by_label`/`by_label` are all keyed by interned `LabelId(u32)`, so both traversal directions resolve `&str → LabelId` once per lookup with zero allocation (pinned by `tests/graph_label_interning.rs` with a counting allocator).

On the format concern this PR's doc comment records — it was right to flag it, and #2171 answers it the way issue #2089 sanctioned: the two String-keyed maps moved to `serde(skip)` + load-time rebuild. Forward compatibility (new code reading pre-change `edge_store.bin`) is pinned by `test_from_bytes_reads_legacy_format_with_trailing_label_maps`, which serializes the old five-field layout and proves it loads with the indices rebuilt; both `load_from_file` and the public `from_bytes` rebuild deterministically. Downgrade (old binary, new file) is explicitly unsupported and its silent load-or-default failure mode is documented in #2171's description.

The nested `node → label → Vec` alternative here and the `(u64, LabelId)` composite in #2171 are allocation-equivalent on the probe path; the composite keeps one map shape for both directions. Since the same lines were rewritten, a rebase would conflict on every hunk for no residual gain. Closing.

---

PR #2190, opened 2026-09-02
fix(quantization): avoid underflow in PQ reconstruct out-of-range error

cyberlife-coder:
## Description

`ProductQuantizer::reconstruct()` builds its `InvalidQuantizerConfig` error
message with `self.codebook.centroids[subspace].len() - 1`. `PQCodebook`
derives `Deserialize` and is never re-validated after load, unlike
`train()`, which does enforce `num_centroids > 0`. If a codebook is
deserialized with an empty centroid table for a subspace, the range guard
(`code_idx >= centroids[subspace].len()`) is always true, so the branch
always executes `0usize - 1` — an integer-underflow panic in a
debug/test build (overflow checks are on by default) or a wraparound to
`usize::MAX` in release, either way defeating the function's documented
`# Errors` contract of returning `Err` on an out-of-range code.

Fix: use `saturating_sub(1)` for the message operand — a one-line change
on the already-erroring path, no behavior change on any success path.

Found via a routine code-quality review of `velesdb-core`; not tied to an
existing issue.

## Type of Change

## How Has This Been Tested?

Added `reconstruct_reports_range_error_without_overflow_on_empty_subspace`,
which builds a `PQCodebook` directly with an empty centroid table and
asserts `reconstruct()` returns `Err(InvalidQuantizerConfig)` rather than
panicking. Verified it reproduces the panic (`attempt to subtract with
overflow`) against the pre-fix code, and passes with the fix.

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing

**Test Configuration:**
- OS: Linux (CI container)

Also ran, all green:
- `cargo clippy -p velesdb-core --lib --bins --features persistence,gpu,update-check -- -A warnings -D clippy::undocumented_unsafe_blocks`
- `cargo clippy -p velesdb-core --all-targets --features persistence,gpu,update-check -- -D warnings -D clippy::pedantic`
- `cargo test -p velesdb-core --features persistence -- --test-threads=1` (5472 passed, 0 failed)
- `cargo test --doc --package velesdb-core`
- `python3 scripts/check_prod_unwraps.py`
- `python3 scripts/check-ai-attribution.py "origin/develop..HEAD"`

## Checklist

- [ ] I have made corresponding changes to the documentation

## Additional Notes

Scope is intentionally surgical (single-line fix + one regression test) per
this repo's working principles — no adjacent refactoring.

### DIFF
--- a/crates/velesdb-core/src/quantization/pq.rs
+++ b/crates/velesdb-core/src/quantization/pq.rs
@@ -234,7 +234,7 @@ impl ProductQuantizer {
                 return Err(Error::InvalidQuantizerConfig(format!(
                     "code index {code_idx} out of range for subspace {subspace} \
                      (max {})",
-                    self.codebook.centroids[subspace].len() - 1
+                    self.codebook.centroids[subspace].len().saturating_sub(1)
                 )));
             }
             let centroid = &self.codebook.centroids[subspace][code_idx];
--- a/crates/velesdb-core/src/quantization/pq_tests.rs
+++ b/crates/velesdb-core/src/quantization/pq_tests.rs
@@ -258,6 +258,31 @@ fn reconstruct_invalid_codes_returns_error() {
 }
 
 #[test]
+fn reconstruct_reports_range_error_without_overflow_on_empty_subspace() {
+    // A deserialized `PQCodebook` is never re-validated against `train()`'s
+    // non-zero-centroid invariant, so `reconstruct` must not underflow while
+    // formatting the "max" operand of its own out-of-range error.
+    use super::PQCodebook;
+
+    let codebook = PQCodebook {
+        centroids: vec![vec![]],
+        dimension: 2,
+        num_subspaces: 1,
+        num_centroids: 0,
+        subspace_dim: 2,
+    };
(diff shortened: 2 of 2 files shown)


cyberlife-coder (commented):
Review.

The diagnosis holds: `PQCodebook` derives `Deserialize` and nothing re-validates `num_centroids > 0` after load, so an empty centroid table makes the range guard unconditionally true and the message operand `len() - 1` underflows — a panic under overflow checks, `usize::MAX` in release, either way breaking the documented `# Errors` contract on the one path that exists to return `Err`.

`saturating_sub(1)` is the right fix and the right size: it touches only the operand of an error message on the already-erroring branch, so no success path changes. The regression test constructs the exact deserialized shape (`centroids: vec[vec[]]`, `num_centroids: 0`) rather than going through `train()`, which is what makes it reach the branch at all; it was shown to reproduce the panic pre-fix.

Scope deliberately not widened to re-validating codebooks at load — that would be a behaviour change on `open` and belongs in its own decision, not this one-liner.

`CI Success` green on `2f0816f7`, 5472 core tests passing. Merging.

---

PR #2191, opened 2026-09-02
feat(guards): refuse a partial Arc/Box trait forwarding (#1967)

cyberlife-coder:

`feat/1967-trait-forwarding-guard` → **`develop`**. ✅

## Description

`velesdb-memory`'s doctrine in `src/lib.rs` already states the rule:

> an adapter over one of these traits forwards the **whole** trait — partial
> forwarding is how a binding silently loses a capability the server already
> publishes (the #1690–#1692 gap family), **and is rejected in review**.

"Rejected in review" is a person paying attention. #1967 item 2 asks for the
machine instead — *« Vérifier qu'aucun trait n'a un forwarding `Arc`/`Box`
partiel (le bug de #1692 en germe) — un test ou un grep-garde si un troisième
trait forwardé apparaît »*. There are five adapters now.

### Why a reviewer misses this

rustc refuses an `impl` that omits a **required** method. So a partial
forwarding can only ever drop a method that has a **default body** — and then
it compiles. `Arc<Concrete>` keeps working and silently runs the trait's
default instead of `Concrete`'s override.

`Extractor::extract_graph` is the shape that already cost this repo a bug:
forward only `extract`, and every `Arc`-held backend loses the relations and
attributes it actually produced. Nothing in the type system says a word.

### What the guard does

`scripts/check-trait-forwarding.py` reads every adapter under `crates/*/src`
(test files excluded) that exists *only to delegate*, and demands it define
every method the trait declares. Two shapes:

```rust
impl<T: Trait + ?Sized> Trait for Box<T> | Arc<T> | Rc<T>   // generic wrapper
impl Trait for DynTrait                                     // erased alias
```

The second shape is the one worth arguing for. `type DynReranker = Box<dyn Reranker + Send + Sync>`
with `impl Reranker for DynReranker` carries **no generic parameter**, so a
wrapper-only pattern never sees it — and it is what the Node binding wraps a
JS callback in. A default-bodied method dropped there is lost for every JS
caller while the Rust tests, which use the concrete type, stay green. That is
the #1690–#1692 failure mode with the serial numbers filed off, and the first
draft of this guard was blind to it.

Five adapters are checked today — four generic wrappers (`Reranker`,
`Embedder`, `Extractor`, `TokenEstimator`) and `DynReranker` — and all five
are whole, so this lands green and stays a tripwire.

Three scoping decisions, all deliberate:

- **A plain `impl Trait for Backend` is exempt.** `impl Extractor for
  OutlineExtractor` does not forward `extract_graph`, and must not be made to:
  a concrete backend is entitled to the trait's default. Only an adapter whose
  entire reason to exist is delegation is held to completeness.
- **Supertrait methods are not demanded.** The storage facets of #1959 are
  separate traits and the doctrine's unit is the facet: *an adapter picks
  which facets it serves, but each facet it implements is forwarded whole.*
- **A trait defined outside the scanned tree is reported `SKIPPED`, not
  passed.** A miss the guard cannot see must never read as a clean run.

### Two parsing details worth the review

Both are places where a shortcut costs *reach*, silently — which is the one
thing a guard must never do.

**Impl headers are closed by matching angles, not by a regex.**
A parameter list nests (`impl<K, T: Store<K> + ?Sized>`), and an `Fn` bound
puts an arrow inside it (`impl<F: Fn(&str) -> u8, ...>`) whose `>` closes
nothing. A `[^>]*` pattern matches neither, so the guard would police fewer
adapters every time a bound grew a generic argument, and report a smaller
number without complaining. Four header shapes are pinned by tests.

**Lifetimes are not char literals.** Rust spends `'` on lifetimes far more
often than on char literals, and a `'static` bound leaves the tick
unmatched. Read as an opening quote, the scan runs past every comment that
follows — so a method named only in prose enters the trait's method set and
the guard demands a forwarding for something that does not exist. A guard
that invents work is a guard someone switches off. `'` therefore opens a
literal only when it actually closes as one; otherwise it is a lifetime and
only the tick is stepped over. Pinned by a test that fails on the pre-fix
lexer.

### One stale line comes with it

`crates/velesdb-memory/src/service.rs`'s module doc still called the
working-context bridge *"the next named cut (#1967, decided together with the
store facetting of #1959)"*. That cut is done — the bridge lives in
`context/memory_bridge.rs` (736 lines) and `context/memory_bridge_compile.rs`
(860), and no working-context code remains in `service.rs`. Merging this PR
without touching that sentence would leave the code pointing at a closed issue
as future work. The replacement is the same **six** lines, so `service.rs`
stays at its baselined 1006 and the file-budget guard neither grows nor
shrinks.

Fixes # (partially — closes item 2 of #1967; items 1, 3, 4 and 5 were
delivered earlier, see the issue comment)

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)

The only Rust touched is a six-line module doc comment. Zero runtime impact:
this PR adds a CI guard, its self-test, its registry entry and its wiring.

## How Has This Been Tested?

- [x] Unit tests
- [x] Manual testing

```
python3 -m unittest scripts.tests.test_check_trait_forwarding   → Ran 19 tests, OK
python3 -m unittest scripts.tests.test_ci_gate_reachability     → Ran 78 tests, OK
python3 -m unittest scripts.tests.test_guard_refusal_vectors    → Ran  8 tests, OK
python3 scripts/check-trait-forwarding.py                       → PASSED: 5 adapters checked
cargo check -p velesdb-memory --features persistence,context    → clean
python3 scripts/check-file-budgets.py                           → PASSED
python3 scripts/check-ai-attribution.py --tree                  → PASSED
ruff check scripts/check-trait-forwarding.py …                  → All checks passed
```

The refusal was proven, not assumed: reconstructing the #1692 bug in a
fixture (`Extractor` for `Arc<T>` forwarding only `extract`) yields

```
crates/velesdb-memory/src/extract.rs:11: `impl Extractor for Arc<T>` forwards 1/2 methods — missing `extract_graph`.
```

with exit 1. The positive control also checks that the scan actually *reaches*
its five adapters — and that `DynReranker` is among them by name, so a
regression that quietly narrowed the guard back to generic wrappers would fail
rather than pass with a smaller number.

**Test Configuration:**
- OS: Ubuntu 24.04 (x86_64)

## Checklist

- [x] I have made corresponding changes to the documentation (CHANGELOG)

## Unsafe Code Checklist

## High-Risk Change Checklist

Not applicable — no `hnsw`, `storage`, `Drop`, `unsafe` or SIMD path is touched.

## Additional Notes

Registry contract per AGENTS.md rule 12 is implemented in full:

- `guards.json` entry — `required: true`, `mode: strict`, job `lint`;
- **declared blind spot**: *presence, not delegation* — a forwarded method
  whose body reimplements the default instead of calling `(**self)` still
  satisfies the guard, and only review catches that. Three further limits are
  named there, each currently vacuous under `crates/*/src`;
- **three `must_refuse` vectors**, each carrying all four fields — the #1692
  shape; a partial forwarding whose trait lives in a *different file* so the
  miss is invisible at the impl site; and a partial forwarding on an erased
  `Dyn*` alias. All three were run by hand through the harness's own helpers
  before pushing, in both states.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -448,6 +448,14 @@ jobs:
       - name: Check durability barriers on file-creating write paths
         run: python3 scripts/check-durability-barrier.py
 
+      # velesdb-memory's doctrine (crates/velesdb-memory/src/lib.rs) requires a
+      # `Box`/`Arc` adapter to forward the WHOLE trait. rustc only demands the
+      # *required* methods, so a partial forwarding can only ever drop one with
+      # a default body — it compiles, and the wrapper silently runs the default
+      # instead of the concrete override. That is the #1690-#1692 gap family.
+      - name: Check wrapper adapters forward their trait whole
+        run: python3 scripts/check-trait-forwarding.py
+
       - name: Check for .unwrap() in production code
         run: python3 scripts/check_prod_unwraps.py
 
@@ -469,6 +477,9 @@ jobs:
 
       - name: Test file-budgets audit script
         run: python3 -m unittest scripts.tests.test_check_file_budgets
+
+      - name: Test trait-forwarding audit script
+        run: python3 -m unittest scripts.tests.test_check_trait_forwarding
 
       - name: Audit feature claims vs exports
         run: python3 scripts/check-feature-claims.py
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,44 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **A guard that no adapter forwards its trait partially (#1967).**
+  `scripts/check-trait-forwarding.py` reads every adapter under `crates/*/src`
+  that exists only to delegate — the generic wrapper
+  `impl<T: Trait + ?Sized> Trait for Box<T> | Arc<T> | Rc<T>`, and the erased
+  alias `impl Trait for DynTrait` where `type DynTrait = Box<dyn Trait>` —
+  and demands the impl define every method the trait declares. An ordinary
+  `impl Trait for Backend` is deliberately exempt: a concrete backend is
+  entitled to the trait's default, which is what a default is for.
+
+  The bug it prevents is narrow, which is exactly why it needs a machine.
+  rustc already refuses an `impl` that omits a *required* method, so a partial
+  forwarding can only ever drop a method that has a **default body** — and
+  then it compiles: `Arc<Concrete>` silently runs the trait's default instead
+  of `Concrete`'s override. `Extractor::extract_graph` is the shape that
+  already cost the repo a bug (the #1690–#1692 gap family): forward only
+  `extract`, and every `Arc`-held backend loses the relations and attributes
+  it actually produced. `velesdb-memory`'s doctrine in `src/lib.rs` states the
+  rule — "an adapter forwards the **whole** trait" — and until now only review
+  enforced it.
+
+  Five forwardings are checked today — four generic wrappers (`Reranker`,
+  `Embedder`, `Extractor`, `TokenEstimator`) and one erased alias
+  (`DynReranker`, which is what the Node binding wraps a JS callback in);
+  all five are whole. The alias shape carries no generic parameter, so a
+  wrapper-only pattern never sees it, and it is precisely where a dropped
+  method would be lost for every JS caller while the Rust tests, which use the
+  concrete type, stayed green.
+
+  Supertrait methods are deliberately not demanded: the storage facets of
+  #1959 are separate traits, and the doctrine's unit is the facet. A trait
+  defined outside the scanned tree is reported `SKIPPED` rather than passed,
+  so a miss the guard cannot see never reads as a clean run.
+
+  Wired into `ci.yml`'s `lint` job with its self-test, registered in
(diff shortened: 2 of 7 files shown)


cyberlife-coder (commented):
Review — read the guard end to end against what would make it lie in either direction.

**Reach (would it silently police less?)** No. Impl headers close by balanced angles with `->` stepped over, so a nested bound or an `Fn` arrow can't shrink the match; the positive control asserts ≥ 5 adapters *and* names `DynReranker`, so a regression back to wrapper-only fails loudly instead of passing with a smaller number. Refusal proven on the #1692 shape, cross-file, and the erased alias — all three vectors run through the harness's own helpers in both states.

**Noise (would it invent work?)** No. A `'static` bound no longer leaks prose into the method set; `fn` inside a string, a comment, or a nested body is excluded; a concrete `impl Trait for Backend` is exempt on purpose. `top_level_fns` matches `fn` regardless of `async`/`unsafe`/`const` prefix, so a trait using those is counted the same as its forwarding. Associated consts and types are not methods and are correctly not demanded.

**Blind spot is declared, not hidden**: presence-not-delegation, `#[cfg]`-gated methods, macro-generated impls, foreign traits reported `SKIPPED`.

Registry contract complete (`must_refuse` ×3 with all four fields, self-test in the required `lint` job). `CI Success` green on `85416069`. Nothing to change. Merging.

---

PR #2192, opened 2026-09-02
fix(bench): let fused_recall_benchmark run under the #1743 degree cap

cyberlife-coder:

`fix/1742-fused-recall-bench-under-degree-cap` → **`develop`**. ✅

## Description

`fused_recall_benchmark` has been unable to run past its first parameter point since #1743 landed, and nothing noticed.

### What was wrong

The bench hung every fact off **one** entity hub and asserted the 2-hop walk reached all of them — a liveness gate so a search regression can't silently benchmark a no-op. Then #1743 capped a single node's expansion at `MAX_WHY_NODE_DEGREE` (64) and the whole walk at `MAX_WHY_NODES` (500). At degree 200 the walk reaches **66** nodes (64 + seed + hub), the assertion panics, and the remaining four of five measurements are never taken:

```
thread 'main' panicked at fused_recall_benchmark.rs:59:
assertion `left == right` failed: traversal did not reach the full hub fan-out
  left: 66
 right: 202
```

So the empirical check for #1742's O(edges + nodes) traversal fix has been dead since the cap. CI didn't catch it because `Internal Bench Compiles` runs `cargo check -p velesdb-core --benches` — only `velesdb-core`. This crate's benches were never built *or* run in CI.

### The fix — two commits

**1. The fixture is now a forest of hubs**, each mentioning at most `MAX_WHY_NODE_DEGREE` facts. The reach the walk covers still grows into the hundreds while every node stays inside the policy the walk actually runs under. The sweep covers both regimes the caps create:

- **Under `MAX_WHY_NODES`** (50, 200, 480): the walk must reach seed + hubs + every fact *exactly*, and must not report truncation.
- **Past it** (1000): the walk must stop at precisely `MAX_WHY_NODES`, must report `truncated`, and the cost must plateau — the cap is a cost bound, and that claim is now checked rather than assumed.

The cheaper fix — relax the assertion to `min(degree, 64) + 2` — was rejected. Past 64 the graph contribution is constant, so the bench would run and measure nothing it was built to measure: a dead guard that looks alive.

**2. CI now type-checks this crate's benches.** One `cargo check -p velesdb-memory --benches --features persistence,context` step beside the `velesdb-core` one in `Internal Bench Compiles`. It's the cheapest reach that catches the compile half of this class on every PR. Deliberately a type-check, not a run: running benches per PR is a cost/coverage policy decision — the trade-off `feature-powerset.yml` already documents for its own sweep — and is not taken here.

### What it measures now

Release, shortened criterion settings (`--measurement-time 5 --sample-size 20`), shared container:

| reach | time | per reached fact |
|---|---|---|
| 50 | 1.17 ms | 23.3 µs |
| 200 | 2.20 ms | 11.0 µs |
| 480 | 4.12 ms | 8.6 µs |
| 1000 (truncated at 500) | 5.04 ms | plateau |

Under the ceiling `t ≈ 0.82 ms + 6.9 µs × reach` fits all three points (predicts 2.19 ms at 200; measured 2.20) — linear, as #1742 promised. Past it the reach doubles and time rises 22%; the residual is vector search over a store twice the size, not traversal.

These are indicative single-run numbers, not a published claim — they show the bench works, and do not go into `BENCHMARKS.md` yet.

Related: #1742 (the fix this bench checks), #1743 (the cap that killed it).

## Type of Change

Bench, one CI step, CHANGELOG. No production code, no runtime impact.

## How Has This Been Tested?

- [x] Manual testing

```
cargo bench -p velesdb-memory --bench fused_recall_benchmark --features persistence
  → all 4 points run, liveness gate holds at each, exit 0   (was: panic at point 2)
cargo check -p velesdb-memory --benches --features persistence,context   → clean (the new CI step, run locally)
cargo clippy -p velesdb-memory --benches --features persistence -- -D warnings  → clean
cargo fmt -p velesdb-memory -- --check                                          → clean
python3 -m unittest scripts.tests.test_ci_gate_reachability                     → 78 tests, OK
python3 scripts/check-ai-attribution.py --tree                                  → PASSED
```

Reproduced the original failure first on an unmodified checkout (the panic above), then the same command passing on this branch.

**Test Configuration:**
- OS: Ubuntu 24.04 (x86_64)

## Checklist

- [x] I have made corresponding changes to the documentation (CHANGELOG)

## Unsafe Code Checklist

## High-Risk Change Checklist

Not applicable — a benchmark file, a CI step and a CHANGELOG entry.

## Additional Notes

Whether to *run* this crate's benches in CI (per PR, or a weekly sweep like the feature powerset) is left open on purpose. It's a policy call with a real cost, and it deserves its own issue rather than a side-effect of a bench fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -740,6 +740,16 @@ jobs:
       - name: Library and benches under internal-bench
         run: cargo check -p velesdb-core --benches --features persistence,internal-bench
 
+      # velesdb-memory's four criterion benches were built by nothing in CI,
+      # so one of them stayed dead for weeks after #1743 changed the policy
+      # it asserted on (its liveness gate panicked at the second parameter
+      # point). Type-checking them here is the cheapest reach that would
+      # have caught the compile half of that; running them stays a policy
+      # decision (cost per PR vs a weekly sweep, the feature-powerset
+      # trade-off) and is deliberately not taken in this job.
+      - name: velesdb-memory benches type-check
+        run: cargo check -p velesdb-memory --benches --features persistence,context
+
       - name: Tests under internal-bench
         run: cargo check -p velesdb-core --tests --features persistence,internal-bench
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`fused_recall_benchmark` could not run past its first parameter point.**
+  It hung every fact off one entity hub and asserted the walk reached all of
+  them — the liveness gate that proves a search regression cannot benchmark a
+  no-op. Issue #1743 then capped a single node's expansion at
+  `MAX_WHY_NODE_DEGREE` (64) and the whole walk at `MAX_WHY_NODES` (500), so
+  at degree 200 the walk reached 66 nodes (64 + seed + hub), the assertion
+  panicked, and four of the bench's five measurements were never taken. The
+  empirical check for #1742's O(edges + nodes) traversal fix had been dead
+  since that cap landed; nothing noticed because CI's `Internal Bench
+  Compiles` only `cargo check`s `velesdb-core`'s benches.
+
+  The fixture is now a forest of hubs, each mentioning at most
+  `MAX_WHY_NODE_DEGREE` facts, so the reach the walk covers still grows into
+  the hundreds while every node stays inside the policy it actually runs
+  under. The sweep covers both regimes the caps create: three points under
+  `MAX_WHY_NODES` where the walk must reach every fact exactly, and one past
+  it where it must truncate at precisely the ceiling and the cost must
+  plateau. Relaxing the assertion to `min(degree, 64)` was rejected — past
+  64 the graph contribution is constant, and a bench that runs but no longer
+  stresses what it was built to stress is a dead guard that looks alive.
+
 - **`VectorCollection::create_with_hnsw` documented a false equivalence.** It
   claimed that passing `None` for both arguments was "equivalent to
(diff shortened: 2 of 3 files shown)


cyberlife-coder (commented):
Review.

**Correctness of the fixture.** `hubs_for(reach) = ceil(reach / 64)` hubs, each with ≤ 64 `mentions`, seed → every hub: at reach 480 the walk visits 1 + 8 + 480 = 489 < 500, so the under-ceiling liveness gate is exact and truncation must be `false`; at 1000 it must stop at exactly `MAX_WHY_NODES` and report `truncated`. Both asserted, both held on the run (all four points, exit 0) where the old bench panicked at point two. The original failure was reproduced first on an unmodified checkout.

**What it now proves.** Under the ceiling, `t ≈ 0.82 ms + 6.9 µs × reach` fits all three points — linear, which is the #1742 claim. Past it, time rises 22 % while reach doubles; that residual is vector search over a store twice the size, not traversal. The cheaper fix (relax the assertion to `min(degree, 64) + 2`) was correctly rejected — it would leave a bench that runs and measures nothing it was built for.

**One nit, not blocking:** at the over-ceiling point `Throughput::Elements(reach)` reports elements = 1000 while the walk reaches 500, so criterion's `thrpt` line overstates by 2× on that row. The `time` line is the measurement and the doc says "plateau", so nothing reads wrong; if this bench ever gains a second over-ceiling point it should report `min(reach, MAX_WHY_NODES)` instead.

**CI step.** `cargo check -p velesdb-memory --benches --features persistence,context` in `Internal Bench Compiles` — passed on this head. Type-check, not run, by explicit decision; the run-in-CI policy is left for its own issue. Correct scope.

Approving on content. Merging once Tests and Lint finish on `9e76f583`.

---

PR #2193, opened 2026-09-02
fix(memory): bound the working-context index at 1000 sessions per project

cyberlife-coder:

`fix/working-index-bounded-per-project` → **`develop`**. ✅

## Description

Two findings from the lock / resilience / memory audit of `velesdb-memory` requested this morning, both in `save_working_context`. The audit's other findings were negative — every sync lock is `parking_lot`, no nested holds, the global index lock is never held across the embedder, every other collection is already capped.

### 1. The working-context index was unbounded

A project's index is **one fact**, rewritten whole on every `save_working_context`: serialised, embedded, stored. It gained an entry per distinct session id and shed one only when that session's fact was forgotten. So:

- a long-lived project paid **O(sessions) per save**, forever;
- `write_working_index` never checked the 1 MiB fact cap — that check (`memory_bridge.rs:214`) guards the working-context *content*, not the index — so nothing bounded the fact's growth at all.

Not a crash. A curve — the kind that never shows up in a test and always shows up in production a year later.

**The bound.** `MAX_WORKING_SESSIONS_PER_PROJECT = 1_000`. Past it, the oldest-saved entries leave the listing. Their facts **stay on disk** and `load_working_context` still finds them by exact `project` + `session` — the semantic an entry a torn index lost already had, so the cap introduces no new failure mode, only a ceiling on an existing one. Survivors keep the read path's order (`saved_at` desc, then session asc).

**The session being saved is pinned before the truncate.** Without that, a same-second flood of other sessions ties on `saved_at`, name order decides, and the very entry this call put there could be evicted — `save_working_context` would report success for a session the listing then cannot see. A unit test builds precisely that tie.

No new lock, no new map, no per-project state. Eviction runs inside `update_working_index`, under the lock it already holds, after `live_sessions` has shed the forgotten entries. A `HashMap` keyed by caller-supplied project names was the obvious alternative and is the unbounded slow leak this same file already declined once, for the write lock, in writing.

### 2. An empty `project` or `session` was accepted

Both are the id the state is filed under and listed by. An empty or whitespace-only one hashed, encoded and listed without complaint — as `""` — and was then unrecoverable by anyone who did not think to ask for the empty string. Now refused before anything is written, with `MemoryError::EmptyWorkingContextKey { key }` naming which one. One `#[non_exhaustive]` variant, one `category()` arm, one check.

## Type of Change

Behavioural: a listing that previously grew forever now holds at most 1 000 entries; an empty key is refused. No API shape change — `mcp-tools.json` is untouched and the MCP doc-contract guard passes. `MCP_TOOLS.md` gains one sentence stating the cap.

## How Has This Been Tested?

- [x] Unit tests

```
cargo test -p velesdb-memory --features persistence,context --lib -- working_index list_working_contexts bound_sessions save_working_context load_working_context refuses_an_empty
  → 22 passed, 0 failed (8.84 s)
cargo clippy -p velesdb-memory --features persistence,context --tests -- -D warnings   → clean
cargo fmt -p velesdb-memory -- --check                                                 → clean
python3 scripts/check-file-budgets.py                                                  → PASSED
python3 scripts/check-doc-freshness.py                                                 → passed
python3 scripts/check-mcp-doc-contract.py                                              → passed
python3 scripts/check-ai-attribution.py --tree                                         → PASSED
```

Three new tests, all pure or near-instant: `bound_sessions` on the same-second tie with the just-saved session sorting last; oldest-first eviction plus the under-cap no-op; and the empty-key refusal for both keys, whitespace included, asserting nothing was written under either. A first draft used a 1 001-save integration test; it proved the same invariants in **88 s** and was replaced — a unit suite should not pay a minute for a `truncate`.

**Test Configuration:**
- OS: Ubuntu 24.04 (x86_64)

## Checklist

- [x] I have made corresponding changes to the documentation (CHANGELOG, MCP_TOOLS.md, model doc)

## Unsafe Code Checklist

## High-Risk Change Checklist

Not applicable — no `hnsw`, `storage`, `Drop`, `unsafe` or SIMD path. The one lock touched is the existing `WORKING_INDEX_WRITE`, and only by adding a `truncate` inside its already-held critical section.

## Additional Notes

Why 1 000: the listing's only consumer is an agent deciding what to resume, which wants the most recent few. A thousand is two orders of magnitude above that, and small enough that the per-save rewrite stays under 100 KB of JSON at typical session-id lengths. If a real workload ever needs more, the constant is one number in `limits.rs` with its reasoning beside it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `VectorCollection::create_with_hnsw_params`,
   `GraphCollection::create_with_hnsw_params`.
 
+### Changed
+
+- **A project's working-context index is bounded at 1 000 sessions
+  (`MAX_WORKING_SESSIONS_PER_PROJECT`).** The index is one fact per project,
+  rewritten whole on every `save_working_context` — serialised, embedded,
+  stored. It gained an entry per distinct session id and shed one only when
+  that session's fact was forgotten, so a long-lived project paid
+  O(sessions) per save, forever, and `write_working_index` never checked the
+  1 MiB fact cap (that check guards the working-context *content*, not the
+  index). Past the cap the oldest-saved entries now leave the listing; their
+  facts stay on disk and `load_working_context` still finds them by exact
+  `project` + `session` — the semantic a torn index already had. The session
+  being saved is pinned and can never be the one evicted, even in a
+  same-second flood of other sessions. Found by the lock/resilience/memory
+  audit of `velesdb-memory`; the audit's other findings were negative (no
+  `std::sync` locks, no nested holds, the global index lock never held across
+  the embedder, every other collection already capped).
+
 ### Fixed
 
 - **`VectorCollection::create_with_hnsw` documented a false equivalence.** It
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`save_working_context` refuses an empty `project` or `session`.** Both
+  are the id the state is filed under and listed by; an empty one hashed,
+  encoded and listed fine — as `""` — and was unrecoverable by anyone who did
+  not think to ask for the empty string. Now
+  `MemoryError::EmptyWorkingContextKey { key }` names which one, before
+  anything is written. Same audit, same file as the index bound above.
+
 - **`VectorCollection::create_with_hnsw` documented a false equivalence.** It
   claimed that passing `None` for both arguments was "equivalent to
   `VectorCollection::create`". It is not: it persists
--- a/crates/velesdb-memory/src/context/memory_bridge.rs
+++ b/crates/velesdb-memory/src/context/memory_bridge.rs
@@ -216,8 +216,11 @@ impl<E: Embedder, S: FactStore> MemoryService<E, S> {
     /// able to cause one on a call that carries nothing (issue #1654).
     ///
     /// # Errors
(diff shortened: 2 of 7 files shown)


cyberlife-coder (commented):
Review.

**`bound_sessions` is right at the edges.** `swap_remove` pulls the pinned entry out before the sort, so the tie-break can never touch it; the rest sort `saved_at` desc then `session` asc — the read path's order, so what survives is exactly what `list_working_contexts` would have shown first. Truncate to `CAP − 1` when pinned, `CAP` otherwise; if the just-saved id is somehow absent, `pinned` is `None` and nothing is lost. Under the cap it returns before touching order, and a test pins that.

**The pin matters, and the test builds the exact case:** CAP + 1 entries sharing one `saved_at`, the just-saved session sorting last by name. Without the pin, `save_working_context` would return success for an entry the listing then couldn't see.

**Eviction runs under the lock already held**, after `live_sessions` has shed the forgotten ones — no new lock, no new map, no per-project state. The `HashMap`-keyed-by-caller-string alternative is the unbounded slow leak this same file already declined in writing for the write lock.

**Empty-key refusal** is before any write, names which key, and the test asserts nothing landed under either `"veles"` or `""`. Only `save` refuses — a read of `""` just returns empty — which is the right asymmetry: the bug was an unrecoverable *write*. `EmptyWorkingContextKey` is `#[cfg(feature = "context")]` in both the variant and its `category()` arm, and the MCP doc-contract guard passes with `mcp-tools.json` untouched.

Semantics stated honestly: past the cap the listing forgets, the store does not, and `load_working_context` by exact id still works — the same semantic a torn index already had, so the cap adds a ceiling to an existing failure mode rather than a new one. `MCP_TOOLS.md` says so in one sentence.

Approving on content. Merging last of the three, after refreshing on `develop` (it shares the `### Fixed` insertion point with #2192).

---

PR #2205, opened 2026-09-04
docs(memory): fix every broken intra-doc link and gate them in CI

cyberlife-coder:

`docs/*` → targets `develop`. ✅

## Description

`velesdb-memory` is published to crates.io, so its rustdoc **is** what docs.rs serves: a broken intra-doc link is a dead link on a public page. **Forty-nine had accumulated** — because nothing in CI ever built the docs.

Two causes, one per class.

**Twenty-two "unresolved".** A module's `//!` header is merged into the `mod` declaration that pulls it in, and once that declaration also carries `///` docs the merged text resolves in the **parent's** scope. So every `super::` path inside the header points one level too high, and every bare private name falls out of scope. The tell: `super::chunk` and `super::chunk_text` were **correct as written** — `chunk_text` really is re-exported at `context::chunk_text` — and both dangled anyway. These become absolute `crate::` paths, which resolve identically in either scope.

**Twenty-seven "links to private item".** Public documentation linking to a private item renders as a link the reader cannot follow. These stop being links: the prose still names `GenerationGate`, `is_reserved_key` or `store_fact` in a code span — it just no longer promises a page that 404s.

**The gate.** The Lint job now runs `RUSTDOCFLAGS=-D warnings cargo doc -p velesdb-memory --no-deps --all-features`. Without it this fix decays: 49 links accumulated precisely because nothing looked. `AGENTS.md`'s pre-push block gains the same command.

Scoped to `velesdb-memory` deliberately — it is the crate whose docs are published from this repo, and I have not audited the others. Widening the gate is a separate call.

## It also changes four MCP tool descriptions

I first wrote that no prose changed. **That was wrong**, and CI caught it: MCP tool descriptions are derived from the doc comments, so `mcp_tools_drift` failed on the `Lint & Format` job. Four descriptions lose their link brackets:

| before | after |
|---|---|
| `` [`super::wire::deserialize_optional_id`] `` | `` `wire::deserialize_optional_id` `` |
| `` [`super::chunk::fence_segments`] `` | `` `chunk::fence_segments` `` |
| `` [`super::classify::classify`] `` | `` `classify::classify` `` |
| `` [`MIN_LOG_RUN_LINES`] `` | `` `MIN_LOG_RUN_LINES` `` |

Every one wrapped a **private** item, so the bracket promised a page that does not exist — inside a description whose reader is a model that cannot look up a Rust path anyway. The wording is otherwise byte-identical. `` [`super::fragment_id`] `` and `` [`CompilePolicy::ids_as_strings`] `` are untouched: their targets are public and resolve. The brackets went only where the target was unreachable.

This is not optional. Those four links are among the ones rustdoc reports, so the gate this branch adds and the old descriptions cannot both stand. `docs/reference/mcp-tools.json` is regenerated with `UPDATE_MCP_TOOLS_SNAPSHOT=1`; `MCP_TOOLS.md` needed no change.

Worth flagging separately: the artifact still carries **18 other `super::`-prefixed Rust paths** in descriptions a model reads. Those resolve in rustdoc, so this change leaves them alone — but they are the same kind of internal noise, and cleaning them up is its own decision.

## Type of Change

- [x] 📚 Documentation update
- [x] 🐛 Bug fix (non-breaking change which fixes an issue) — dead links on a published docs.rs page

## How Has This Been Tested?

- [x] Unit tests
- [x] Integration tests
- [x] Manual testing

Measured, not estimated — the count came from `cargo doc`, iterated in three passes:

| pass | warnings |
|---|---|
| before | 49 |
| after the `crate::` paths | 32 |
| after the private-item links | 8 |
| after the last `Self::`/private ones | **0** |

The gate itself was verified before wiring: `RUSTDOCFLAGS="-D warnings" cargo doc -p velesdb-memory --no-deps --all-features` → **exit 0**. A gate that cannot pass on the branch that adds it is worse than none.

After regenerating the artifact: `cargo test -p velesdb-memory --test mcp_tools_drift` **1 passed**; `check-mcp-doc-contract` OK; `check-doc-freshness` OK; `test_ci_gate_reachability` 97 tests OK; `ci.yml` parses as YAML. The pre-commit hook ran formatting, `cargo clippy --workspace -D warnings`, the `undocumented_unsafe_blocks` pass and the workspace `--lib` suite — all green.

**Why the drift was not caught before pushing:** the pre-commit hook runs `cargo test --workspace --lib`, and `mcp_tools_drift` is an integration test. My own gap, not the hook's — the pre-push block in `AGENTS.md` has `check-mcp-doc-contract.py` and I ran it on the sibling branch but not on this one.

**Test Configuration:**
- OS: Linux x86_64
- Rust version: workspace pinned toolchain

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works — the CI gate *is* the test: it fails on any regression, and it was proven to pass only after the last link was fixed

## Additional Notes

Every broken link was pre-existing; none was introduced by recent work. The `//!`-merge behaviour is worth knowing about when adding module docs — a `super::` path that looks obviously right will still dangle if the `mod` declaration carries its own `///`.

`Gate Contracts / npm advisories (demos/tauri-rag-app)` also went red on an earlier head with a `503 Service Unavailable` from npmjs.org's advisory endpoint; that is an outage, not this diff, and it is written up in a comment below.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -200,6 +200,18 @@ jobs:
             echo "::endgroup::"
           done
 
+      # velesdb-memory is published to crates.io, so its rustdoc IS what docs.rs
+      # serves: a broken intra-doc link is a dead link on a public page. Nothing
+      # in CI built the docs, so 49 of them had accumulated unnoticed — every
+      # `super::` path in a module's `//!` header breaks once the `mod`
+      # declaration also carries `///` docs, because the merged doc resolves in
+      # the parent's scope. Fixed in this change; this step is what keeps them
+      # fixed.
+      - name: Rustdoc intra-doc links (velesdb-memory)
+        env:
+          RUSTDOCFLAGS: -D warnings
+        run: cargo doc -p velesdb-memory --no-deps --all-features
+
       # The extraction backend's unit tests are pure parsing and prompt
       # building: none opens a socket. Run them under the canonical role name
       # and its compatibility alias so neither path can silently stop enabling
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -92,6 +92,9 @@ cargo check -p velesdb-memory --no-default-features --features mcp,persistence
 for feat in mcp http context persistence embedder-http extractor-http ollama extract; do
   cargo check -p velesdb-memory --no-default-features --features "$feat" --all-targets
 done
+# velesdb-memory's rustdoc is what docs.rs serves, so a broken intra-doc link is
+# a dead link on a public page. Blocking in the Lint job.
+RUSTDOCFLAGS="-D warnings" cargo doc -p velesdb-memory --no-deps --all-features
 # Functional wasm gate — catches std APIs that abort on wasm32 (e.g. SystemTime::now):
 wasm-pack test --node crates/velesdb-wasm
 # Guards and contracts. Every one of these blocks a PR through `CI Success`, and
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -93,7 +93,7 @@ for feat in mcp http context persistence embedder-http extractor-http ollama ext
   cargo check -p velesdb-memory --no-default-features --features "$feat" --all-targets
 done
 # velesdb-memory's rustdoc is what docs.rs serves, so a broken intra-doc link is
-# a dead link on a public page. Blocking in the Lint job.
+# a dead link on a public page. Blocking in the Lint job since #2205.
 RUSTDOCFLAGS="-D warnings" cargo doc -p velesdb-memory --no-deps --all-features
 # Functional wasm gate — catches std APIs that abort on wasm32 (e.g. SystemTime::now):
 wasm-pack test --node crates/velesdb-wasm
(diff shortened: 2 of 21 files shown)


cyberlife-coder:
`Gate Contracts / npm advisories (demos/tauri-rag-app)` is red on `08bbecea`, and it is not this pull request's.

The job failed because npmjs.org's advisory endpoint was down:

```
npm warn audit 503 Service Unavailable - POST https://registry.npmjs.org/-/npm/v1/security/advisories/bulk
{ error: 'Service Unavailable' }
npm error audit endpoint returned an error
```

`npm audit` retried for five minutes (00:19:21 → 00:24:21 UTC) before giving up. That is an outage in an external service, not an advisory against a dependency — no package was reported vulnerable, the request never got an answer.

This diff cannot cause it: it changes Rust doc comments, `ci.yml` and `AGENTS.md`, and touches no `package.json`, no lockfile, and nothing under `demos/`.

Corroborated on a second pull request: the same job failed the same way on #2204's earlier head at 00:03 UTC and passed on its next run against identical npm content.

A re-run is refused while the workflow is still going ("This workflow is already running"). I will re-run this job once the run settles, and keep the pull request watched until it is green — no changes will be pushed for this failure.

PR #2207, opened 2026-09-04
docs(core): fix 30 broken rustdoc intra-doc links and deny them at the crate root

cyberlife-coder:

`docs/core-intra-doc-links` → **`develop`**. ✔

## Description

`cargo doc -p velesdb-core` emitted **30 broken-link diagnostics**. Nothing in CI built the crate's docs, so they had accumulated unseen — the same gap #2205 closed for `velesdb-memory`, this time on the crate every binding documents against.

Two distinct defects, two distinct fixes.

**Six links were unresolved.** All six sit in a module's `//!` header, and each broke for the reason #2205 identified: a `//!` header is merged into the `mod` declaration that pulls it in, and once that declaration also carries `///` docs, the merged text resolves in the **parent's** scope. The paths were correct where they were written and dangling where they were read. Fixed with reference-link definitions appended inside each `//!` block, which resolve from the crate root:

| File | Links |
|---|---|
| `conformance/mod.rs` | `check_stable_hash`, `check_rrf`, `check_executor`, `Divergence` |
| `contiguous_file_arena.rs` | `ContiguousVectors` (×2) |
| `observer/context.rs` | `DatabaseObserver` |

**Twenty-four pointed at items a reader of the public docs cannot reach** — `pub(crate)` (`save_config`, `remove_edge_detailed`, `read_gate_raw`, `RawGateOutcome`, `add_column_backfilled`, `distance_from_prepared_slice`), `pub(super)` (`resolve_hnsw_params`, `PREFILTER_RECALL_GUARD`), private modules (`storage::wal_entry`, `log_payload`, `log_payload_io`, `collection::config_serde`) and private items (`decode_ids`, `decode_vectors`, `sync_wal_or_resync`, `HnswIndex::mappings`). Each stops being a link and becomes a plain code span: the prose still names the item, it just no longer promises a page that would 404.

No sentence was rewritten and no claim changed — only link syntax and line breaks.

Two details that are easy to get wrong when demoting a link:

- **Re-wrapping.** Removing `[...]` shortens the line, and `rustfmt` cannot re-flow doc comments (`wrap_comments` is nightly-only and off here). Every affected paragraph is re-wrapped to the surrounding 79 columns rather than left with a 40-column line in the middle of it — including the seams where the edited text meets untouched lines.
- **Keeping the qualification.** ``[`Allow`](RawGateOutcome::Allow)`` becomes `` `RawGateOutcome::Allow` ``, not a bare `` `Allow` ``: a variant name with no type tells the reader nothing.

**The gate is the point of the change, not the cleanup**, and it is in two layers.

The idiomatic layer is the crate root. Both published crates now carry the inner attribute `#[deny(rustdoc::broken_intra_doc_links, rustdoc::private_intra_doc_links)]`.

These lints are **warn-by-default** — which is exactly why 30 links here and 49 in `velesdb-memory` (#2205) accumulated without anything going red. The attribute is what makes a plain `cargo doc` fail, for a contributor locally and for docs.rs, instead of only the one CI step that exports `RUSTDOCFLAGS`. Verified by reintroducing a single private link and watching an *unflagged* `cargo doc` refuse it:

```
error: public documentation for `create_graph_collection_with_embeddings`
       links to private item `Database::resolve_hnsw_params`
error: could not document `velesdb-core`
```

The second layer is CI, and it is not redundant: **an attribute cannot fire if nobody builds the docs.** A new Lint step runs `cargo doc -p velesdb-core --no-deps --all-features`, mirrored into the `AGENTS.md` pre-push sequence. `--all-features` rather than the default `persistence` set is deliberate: docs.rs serves default features only, but a gate blind to the `gpu`, `openapi` and `loom` modules cannot fail on them — the reasoning the isolated-feature step already carries (#1765). It costs the wgpu dependency build in that job.

`velesdb-memory`'s crate root gets the same one-line attribute. Its CI gate already exists (#2205); leaving the two published crates asymmetric is the kind of drift `CLAUDE.md` opens by warning about.

### Why not `[workspace.lints.rustdoc]`

It would be the tidier home, and the workspace already uses `[workspace.lints.clippy]` — but a workspace-wide deny only works if every inheriting crate is clean, so I measured instead of assuming:

| Crate | Broken links | Published to crates.io |
|---|---|---|
| `velesdb-core` | 30 → **0** | yes |
| `velesdb-memory` | 0 (fixed in #2205) | yes |
| `velesdb-cli` | 0 | — |
| `velesdb-mobile` | 0 | — |
| `velesdb-wasm` | 8 | — |
| `tauri-plugin-velesdb` | 4 | — |
| `velesdb-server` | 3 | — |
| `velesdb-migrate` | 2 | — |

17 remain, all in crates that are not published, so their rustdoc is not served by docs.rs. That is the line this PR draws: fix and gate the two crates whose docs are a public artifact, and record the remainder in the changelog as known future work rather than expanding an already 22-file change into a workspace sweep with its own verification burden.

## Type of Change

- [x] 📚 Documentation update

## How Has This Been Tested?

- [x] Unit tests
- [x] Manual testing

```
RUSTDOCFLAGS="-D warnings" cargo doc -p velesdb-core --no-deps --features persistence   # 30 -> 0
RUSTDOCFLAGS="-D warnings" cargo doc -p velesdb-core --no-deps --all-features           # 30 -> 0
RUSTDOCFLAGS="-D warnings" cargo doc -p velesdb-memory --no-deps --all-features         # still clean
```

Both feature sets were verified at 0 **before** the gate was wired, so the step is not being introduced red.

Because a doc comment is a published artifact in this repo, the surfaces that derive from one were re-checked rather than assumed:

- `cargo test -p velesdb-memory --test mcp_tools_drift` — **passes**; the published MCP tool descriptions derive from `velesdb-memory`, not core, so this change leaves `docs/reference/mcp-tools.json` byte-identical. (On #2205 the equivalent change *did* move four descriptions; that is why it is checked and not assumed.)
- `check-mcp-doc-contract.py`, `check-doc-contract.sh`, `check-promise-contract.py`, `check-feature-claims.py`, `check-doc-freshness.py` — all pass.
- `cargo fmt --all -- --check`, `cargo clippy -p velesdb-core --all-targets --features persistence -- -D warnings` — clean. (`clippy` does not object to the `rustdoc::` lint namespace; rustc ignores those lints outside rustdoc.)
- `python3 -m unittest discover -s scripts/tests` — passes.
- The crate-root deny proven to bite, per the mutation above.

**Test Configuration:**
- OS: Ubuntu (x86_64)
- Rust: stable toolchain pinned by `rust-toolchain.toml`

## Checklist

- [x] I have added tests that prove my fix is effective (the crate-root deny and the CI step are the test: either fails on any regression of these 30)

## Unsafe Code Checklist

Not applicable — no `unsafe` added or modified.

## High-Risk Change Checklist

Not applicable — `hnsw`, `storage` and `quantization` files are touched, but only inside doc comments. No executable line changed.

## Additional Notes

The 24 private-target links are demoted rather than made public. Promoting `pub(crate)` items to `pub` to satisfy rustdoc would widen a compatibility surface to fix a documentation defect — the wrong trade in a crate whose public API is a promise.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -212,6 +212,19 @@ jobs:
           RUSTDOCFLAGS: -D warnings
         run: cargo doc -p velesdb-memory --no-deps --all-features
 
+      # Same gate, same reason, for the other published crate: velesdb-core is
+      # the API every downstream binding documents against, and it had carried
+      # 30 broken links (6 unresolved, 24 pointing at `pub(crate)` items) that
+      # nothing in CI looked at. `--all-features` rather than the default
+      # `persistence` set is deliberate: docs.rs serves default features only,
+      # but a gate that skips the `gpu`, `openapi` and `loom` modules cannot
+      # fail on them — the same reasoning as the isolated-feature step above
+      # (#1765). It costs the wgpu dependency build in this job.
+      - name: Rustdoc intra-doc links (velesdb-core)
+        env:
+          RUSTDOCFLAGS: -D warnings
+        run: cargo doc -p velesdb-core --no-deps --all-features
+
       # The extraction backend's unit tests are pure parsing and prompt
       # building: none opens a socket. Run them under the canonical role name
       # and its compatibility alias so neither path can silently stop enabling
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -92,9 +92,13 @@ cargo check -p velesdb-memory --no-default-features --features mcp,persistence
 for feat in mcp http context persistence embedder-http extractor-http ollama extract; do
   cargo check -p velesdb-memory --no-default-features --features "$feat" --all-targets
 done
-# velesdb-memory's rustdoc is what docs.rs serves, so a broken intra-doc link is
-# a dead link on a public page. Blocking in the Lint job since #2205.
+# The published crates' rustdoc is what docs.rs serves, so a broken intra-doc
+# link is a dead link on a public page. Both crates now `#![deny]` the two link
+# lints at their crate root, so a plain `cargo doc` already fails on one; the
+# RUSTDOCFLAGS here and the matching Lint steps are what guarantee the docs are
+# BUILT — an attribute cannot fire if nobody builds them.
 RUSTDOCFLAGS="-D warnings" cargo doc -p velesdb-memory --no-deps --all-features
+RUSTDOCFLAGS="-D warnings" cargo doc -p velesdb-core --no-deps --all-features
 # Functional wasm gate — catches std APIs that abort on wasm32 (e.g. SystemTime::now):
 wasm-pack test --node crates/velesdb-wasm
 # Guards and contracts. Every one of these blocks a PR through `CI Success`, and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **30 broken rustdoc intra-doc links in `velesdb-core`, and the two lints that
+  let them accumulate.** Six were unresolved — all in a module's `//!` header,
+  which rustdoc merges into the `mod` declaration that pulls it in, so once
+  that declaration also carries `///` docs the paths resolve in the *parent's*
+  scope (the same trap as the 49 links fixed in `velesdb-memory`). Twenty-four
+  pointed at `pub(crate)`, `pub(super)` or private items a reader of the
+  published docs cannot reach; those stop being links and become plain code
+  spans rather than promising a page that would 404. No claim changed — only
+  link syntax and the line wrapping that demoting a link disturbs, which
+  `rustfmt` cannot re-flow (`wrap_comments` is nightly-only).
+
+  `rustdoc::broken_intra_doc_links` and `rustdoc::private_intra_doc_links` are
+  warn-by-default, which is why 30 here and 49 there went unnoticed. Both
+  published crates now `#![deny]` them at the crate root, so a plain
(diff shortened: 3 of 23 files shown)


cyberlife-coder:
`Gate Contracts / npm advisories (sdks/typescript)` failed on `6505d4a7`. It is not this PR's failure, and it is not a vulnerability.

The job log is unambiguous — the step started at 10:35:47 and produced this at 10:40:49:

```
npm warn audit 503 Service Unavailable - POST https://registry.npmjs.org/-/npm/v1/security/advisories/bulk
{ error: 'Service Unavailable' }
npm error audit endpoint returned an error
Process completed with exit code 1
```

Five minutes and two seconds of hang, then a 503 from npmjs.org's bulk advisory endpoint, then exit 1. No advisory was found because no audit ever completed. This PR touches only `///` and `//!` doc comments plus one CI step; it changes no lockfile.

This is the same npmjs.org outage that turned this job red three times on `develop` yesterday, and it is exactly the defect #2208 fixes: `npm audit` exits 1 both for an advisory and for an endpoint that never answered, so the job could not tell them apart. Under #2208's gate the same audit is retried with backoff, each attempt bounded, and a persistent transport failure exits 75 with a message saying it is infrastructure rather than a finding. On #2208's own run just now, `npm advisories (examples/react-wasm-search)` completed in 8 seconds.

I have not ported #2208 into this PR: it is a 620-line change across a script, its test suite, the guard registry and the workflow, and duplicating it into a documentation PR would guarantee a conflict between the two. The faster path to the same result is landing #2208 and merging `develop` here, which carries the fix with no duplication.

No re-run spent yet. CI is currently re-running on `75f0f386`; if the endpoint has recovered this job goes green on its own, and if it fails again I will use the single re-run.

PR #2209, opened 2026-09-05
fix(column_store): validate GeoPoint range on update, not just insert

cyberlife-coder:

## Description

`update_by_pk` and `update_multi_by_pk` in `ColumnStore` never validated
`GeoPoint` coordinate ranges, while `insert_row` and the upsert path
(`validate_value_types`) both call `haversine::validate_coordinates`. A
single- or multi-column update could silently write an out-of-range
coordinate (e.g. `GeoPoint(999.0, -500.0)`) that later feeds
`GEO_DISTANCE`/`GEO_BBOX` filtering and the Haversine formula itself.

This adds the same `validate_coordinates` call to `update_by_pk` and to
`validate_multi_update` (used by `update_multi_by_pk`), matching the
existing insert-path behavior. Purely additive: it turns a silent bad
write into the `TypeMismatch` error both functions' contracts already
allow (no signature or persistence-format changes).

Fixes: none filed — found during a routine code-quality review of the
`column_store` CRUD paths.

## Type of Change

## How Has This Been Tested?

- [x] Unit tests

Added two regression tests: `test_update_by_pk_rejects_out_of_range_geopoint`
and `test_update_multi_by_pk_rejects_out_of_range_geopoint`. Ran the full
`column_store` test module (177 tests) with `--features persistence`, plus
`cargo fmt --check` and `cargo clippy --lib --bins --features
persistence,gpu,update-check -- -D warnings -D clippy::pedantic` — all clean.

**Test Configuration:**
- OS: Linux

## Checklist

## Unsafe Code Checklist

N/A — no `unsafe` code added or modified.

## High-Risk Change Checklist

N/A — does not touch `hnsw`, `storage`, `Drop`, `unsafe`, or SIMD dispatch.

## Additional Notes

Scope is deliberately narrow: only the two update paths gain the check
that insert already has. No persistence format, public API, or config
surface changes.

### DIFF
--- a/crates/velesdb-core/src/column_store/batch.rs
+++ b/crates/velesdb-core/src/column_store/batch.rs
@@ -4,6 +4,7 @@
 
 use std::collections::HashMap;
 
+use super::haversine;
 use super::types::{
     BatchUpdate, BatchUpdateResult, BatchUpsertResult, ColumnStoreError, ColumnValue, ExpireResult,
     TypedColumn, UpsertResult,
@@ -79,18 +80,12 @@ impl ColumnStore {
         for (col_name, col_updates) in by_column {
             if let Some(col) = self.columns.get_mut(col_name) {
                 for (row_idx, value) in col_updates {
-                    let actual_type = Self::value_type_name(&value);
-                    if Self::set_column_value(col, row_idx, value).is_ok() {
-                        result.successful += 1;
-                    } else {
-                        let pk = row_to_pk.get(&row_idx).copied().unwrap_or(0);
-                        result.failed.push((
-                            pk,
-                            ColumnStoreError::TypeMismatch {
-                                expected: Self::column_type_name(col),
-                                actual: actual_type,
-                            },
-                        ));
+                    match Self::set_column_value(col, row_idx, value) {
+                        Ok(()) => result.successful += 1,
+                        Err(err) => {
+                            let pk = row_to_pk.get(&row_idx).copied().unwrap_or(0);
+                            result.failed.push((pk, err));
+                        }
                     }
                 }
             } else {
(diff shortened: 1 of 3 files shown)


cyberlife-coder:
Review pass on this branch found a third write path the original shape did not cover, and the fix moved as a result.

## `batch_update` was still writing out-of-range coordinates

`batch_update` goes through `set_column_value` directly (`batch.rs:82`), with neither of the two guards this PR added. So the bug this PR closes stayed open on a third caller — which is the tell that the invariant was being *restated per caller* rather than *held* anywhere.

It now lives in `checked_set_geopoint`, the single primitive every validated write funnels through. Three consequences:

- **`update_by_pk`'s ad-hoc check is gone**, and with it an error-precedence bug it had introduced: validating before `resolve_live_row` meant `update_by_pk(missing_pk, "loc", bad_point)` answered `TypeMismatch` where the rest of the store answers `RowNotFound`, and likewise for an unknown column.
- **`update_multi_by_pk` keeps its pre-check**, because it applies columns one at a time and has to reject the batch before the first write. That is now a comment in the code, so the next reader does not delete it as a duplicate of the primitive's check. It also moved after `validate_type_match`, so a GeoPoint sent to an `Int` column reports the type mismatch rather than a latitude range error.
- **Any future write path is covered** without remembering to add anything.

## `apply_column_updates` was reporting the wrong error

It rewrote every failure as `TypeMismatch { expected: <column type>, actual: <value type> }`. That is byte-for-byte what `set_column_value` already returns for a genuine type mismatch — and a lie for everything else: it reported `expected: GeoPoint, actual: GeoPoint` for a range error, and mislabelled `IndexOutOfBounds` the same way. Propagating the real error is shorter and honest.

## Tests

`assert!(result.is_err())` proved neither which error came back nor that nothing was written. They now pin the error variant, assert the stored point is untouched, cover the multi-column atomicity case (a valid column paired with an invalid one must leave both alone), and cover `batch_update`.

That last one is the load-bearing test, and it was verified red before green: with the guard removed from `checked_set_geopoint` it fails with

```
assertion `left == right` failed
  left: 1
 right: 0
```

`successful: 1` — the write the previous shape let through.

## Local verification

`cargo fmt --check`, 178 `column_store` tests (`--features persistence`), and `cargo clippy --lib --bins --features persistence,gpu,update-check -- -D warnings -D clippy::pedantic`, all clean on `aarch64-apple-darwin`.

One note on the pre-commit hook: it could not run on macOS at all, because `velesdb-node`'s lib test fails to link (`_napi_reference_unref` undefined for arm64). That is unrelated to this PR and is fixed in #2210; the hook sequence above was run with that fix present in the tree so the gate was actually exercised, not skipped.

Issue #1987, opened 2026-08-18
Quality toolchain — standing gates + gaps to wire into CI


cyberlife-coder:
> **Status 2026-09-09 — gaps 1, 2 and 3 are wired; 4, 5, 6 and 7 are not.**
>
> **Gap 2 closed since the last refresh.** `ci.yml` now declares a `semver-checks`
> job running `cargo semver-checks check-release -p velesdb-core --release-type minor`
> and `-p velesdb-memory`. It is wired on *both* halves — present in `ci-success`'s
> `needs` **and** read by its `[[ ... == "success" ]]` chain — so a public-API break
> fails the one required check rather than merely showing red on the PR page.
>
> `--release-type minor` is load-bearing, not a loosening: when the in-tree version
> equals the published one, the tool infers *Patch* and arms all 207 `Deny` lints,
> which would have deadlocked `develop` on the first additive change. `cargo-public-api`
> (the reviewable API diff that pairs with it) is still absent.
>
> **Gap 4 (`cargo-mutants`) is deferred on purpose, not forgotten.** The binary is not
> installed and the workspace holds ~1640 Rust files; mandating a gate whose cost has
> never been measured is how a gate gets switched off. The intended first step is a
> non-blocking `--in-diff` report, so the cost is known before anything blocks.
>
> Absent from all 25 workflows, verified today by grep: `public-api`, `mutants`,
> `minimal-versions`, `cargo-msrv`, `cargo-about`, `geiger`, `codeql`, `semgrep`,
> `oss-fuzz`.
>
> **New item, not in the ranked list below: replaying the gates locally.** A local
> `cargo clippy -p velesdb-core --lib` passed while CI, running `--workspace
> --all-targets`, compiled the test code and refused — a green workstation followed by
> a red public PR. The fix is a script that *reads* the gates from `ci.yml` rather than
> copying them, so the list cannot drift ([#2229](https://github.com/cyberlife-coder/VelesDB/pull/2229)).
> Its two defects worth recording: it first replayed the `lint` job alone and reported
> all-green on a tree CI then failed on `hygiene` and `gate-contracts`, and it reported
> a missing `cargo machete` as a failing gate because it knew only the shell's wording
> for an absent command, not cargo's.
>
> The "already enforced" table below is still accurate (miri/loom/fuzz in
> `quality-deep.yml`, tsan, sonar, llvm-cov), as are the two standing rules.
---

[...]

PR #2217, opened 2026-09-06
ci(quality-gates): add cargo-semver-checks against velesdb-core, velesdb-memory

cyberlife-coder:

`ci/*` → targets `develop`. ✅

## Description

Gap 2 of #1987's quality-toolchain backlog: wire `cargo-semver-checks` as a blocking CI gate, mirroring gaps 1 (#1988) and 3 (#1989), which shipped the same way — a small, self-contained, single-tool CI job.

New job diffs each published library's current public API against its last crates.io release and fails the build if a breaking change ships without the major version bump SemVer requires. Scoped to the two crates this repo treats as consumable Rust libraries: `velesdb-core` (the embeddable engine) and `velesdb-memory` (independent 0.x versioning per `AGENTS.md`). `velesdb-cli` has no `[lib]` target; `velesdb-server`/`velesdb-mobile` are consumed through their own binding surfaces (REST, UniFFI), not as Rust dependencies, so their crate-level API isn't the contract that matters here — left out to keep this PR minimal, and easy to extend later if wanted.

Tool pinned to `0.46.0`, the last release compatible with this workspace's MSRV (1.90) — 0.50.0+ requires rustc 1.93.

**Baseline-before-gating (per `AGENTS.md`'s own standing rule):** ran the tool against the tree before wiring anything. `velesdb-memory` came back clean against `velesdb-memory-v0.14.2`. `velesdb-core` did not, against `v6.0.0`: #2212 turned `bytemuck` from an optional dependency into a mandatory one, which silently dropped the `bytemuck` feature Cargo had derived from `optional = true` (and that `gpu` used to enable transitively). Any downstream `Cargo.toml` still naming that feature — directly or via `gpu` — would fail to resolve. Restored it as a no-op compatibility stub (`bytemuck = []`, `gpu` re-lists it) so the gate starts green instead of red on arrival, the same way #1989 baselined `typos` with an audited allowlist before gating rather than after.

(Local `cargo-semver-checks` runs against this sandbox's registry endpoint hit a TLS trust quirk specific to its network interception, so verification used `--baseline-rev` against the git tags above instead of the registry — that flag isn't in the CI job itself, which uses the tool's normal registry-lookup default and will run against unrestricted GitHub Actions network access.)

Also corrected `QUALITY_BAR.md`'s CI job count, which this PR's own edit to `ci-success`'s `needs:` touches directly. That count was already stale before this change (`memory-feature-matrix`, `internal-bench-check`, `drop-tightening-check` and `hygiene` were absent from both the count and its prose enumeration) — only the number is fixed here; the enumeration is pre-existing drift left for a separate pass.

## Type of Change

- [x] 🔧 Refactoring / tooling (no functional changes to the engine)
- [x] 📚 Documentation update (QUALITY_BAR.md job count)

## How Has This Been Tested?

- `cargo check -p velesdb-core` across `--no-default-features`, default (`persistence`), `--features gpu`, and `--features persistence,gpu,update-check`
- `cargo semver-checks check-release -p velesdb-core --baseline-rev v6.0.0` and `-p velesdb-memory --baseline-rev velesdb-memory-v0.14.2` — both clean after the `bytemuck` fix
- `taplo fmt --check`, `typos crates/ docs/ scripts/ README.md CHANGELOG.md` (the hygiene job's exact scope)
- `python3 -m unittest discover -s scripts/tests -p "test_*.py" -t .` — 810 tests, OK (includes `test_ci_gate_reachability`'s needs/chain-agreement invariant, which would fail if `semver-checks` were added to one list but not the other)
- `check-doc-freshness.py` (all five guards), `check-version-sync.py`, `check-promise-contract.py`, `check-doc-contract.sh`, `check-ai-attribution.py --tree`
- `cargo fmt --all --check`

**Test Configuration:**
- OS: Linux (CI: ubuntu-latest)
- Rust version: 1.90.0 (workspace MSRV)

## Checklist

- [ ] I have added tests that prove my fix is effective — N/A, this is a CI workflow addition plus a Cargo.toml feature-compat fix; `test_ci_gate_reachability.py` already covers the workflow-wiring invariant

## Additional Notes

Not in scope here, left for whoever picks them up next per the same issue's backlog: gap 4 (`cargo-mutants`), gap 5 (`cargo-minimal-versions` + `cargo-msrv`), gap 6 (`cargo-about` + `cargo-geiger`), gap 7 (CodeQL/Semgrep/OSS-Fuzz).


cyberlife-coder:
Adversarial review before merge — **RESERVATIONS**, both majors now fixed on the branch (`2ed32798`). The wiring was correct and the `bytemuck` diagnosis was right; two things would have bitten.

## 1. The compatibility stub restored the name, not the effect

The 6.0.0 manifest carried `features = ["derive"]` on the optional dependency, so enabling the implicit `bytemuck` feature gave the caller bytemuck **with its derive macro**. `bytemuck = []` enables nothing.

A consumer who wrote `velesdb-core = { features = ["bytemuck"] }` alongside their own `bytemuck` and expected `Pod` to be derivable would still break — the same break this PR exists to undo, moved from resolution to compilation.

**The gate cannot catch it.** `feature_missing` and `feature_no_longer_enables_feature` read this package's feature graph, where an empty stub looks restored. A gate that blesses a half-fix is worse than no gate.

Now `bytemuck = ["bytemuck/derive"]`, and `gpu` is byte-for-byte its 6.0.0 line again. Proven rather than argued:

```
$ cargo tree -p velesdb-core -e features --features bytemuck
├── bytemuck feature "default"
│   └── bytemuck v1.25.2
│       └── bytemuck_derive feature "default"
│           └── bytemuck_derive v1.10.2 (proc-macro)
```

All five feature surfaces exit 0: `--no-default-features`, default, `bytemuck`, `gpu`, `persistence,gpu,update-check`.

## 2. The gate was armed at a level `develop` cannot satisfy

`develop` carries the published version between releases — the bump lives in a `chore(release)` commit. cargo-semver-checks reads equal versions as a **patch** update, which satisfies no required level, so all 207 `Deny` lints arm. That includes the 21 needing only a minor bump: `function_must_use_added`, `*_marked_deprecated` and friends.

This crate holds **1369 `#[must_use]`** and deprecates public items as first-class practice. The next PR adding either would have turned `CI Success` red, with no remedy short of a version bump the process performs only at release — and #2219 now locks that further. A deadlock between two gates, neither of them wrong on its own.

`--release-type minor` on core keeps every breaking lint armed and drops the reminder the release commit covers anyway. velesdb-memory needs no flag: a 0.x baseline already reads as minor, so its minor lints are disarmed by the tool. Same enforcement, reached differently — the job now says so rather than leaving it as an unexplained asymmetry.

## Three smaller ones, all documentation that was wrong rather than thin

- The pin comment blamed 0.50.0 for the MSRV ceiling; **0.47.0** is where 1.91 starts. And the MSRV argument does not apply here at all — `install-action` fetches a prebuilt binary, so the pin is really about keeping rustdoc-JSON in step with the toolchain that produces it.
- `QUALITY_BAR.md` said 26 where its own cited authority now holds **27**, of which 26 are read by the chain (`sonarcloud` is listed but deliberately not gating). Counted, not guessed.
- `bytemuck` was missing from the `## Feature flags` enumeration — the same non-exhaustiveness this PR fixes elsewhere.

## One reservation left open, deliberately

The CI invocation uses the tool's default feature heuristic, which activates `gpu`, `loom`, `internal-bench`, `test-fault-injection` and `bench-sift1m` — a set no local check in this PR covered. `cargo doc --all-features` already passes in CI and loom is gated by `cfg(loom)` rather than the feature, so it should hold; but nobody has watched it hold. The first CI run on this branch is that observation.

## What the review confirmed

27 entries in `needs:`, `semver-checks` present in both the list and the chain and compared `== "success"`; no registry declaration required (the guard registry scans `scripts/`, and `hygiene`/`security` are absent for the same reason); actions pinned by SHA; no false green on a network or baseline failure, since a skipped or cancelled job fails `ci-success` rather than passing it; and `bytemuck` is the **only** implicit feature either gated crate lost to this mechanism.

cyberlife-coder:
Both fixes from the review look right to me — the `bytemuck = ["bytemuck/derive"]` correction restores actual compile-time parity, not just resolution, and `--release-type minor` on core is the correct read of what `develop`'s in-between-releases version can honestly claim.

Two follow-ups pushed on top of `2ed32798`:

- Merged `develop` in (`5f9a4929`) — the branch had fallen behind after #2218/#2219 landed; `PR Governance` was red on the freshness check for exactly that reason, not the content of either commit.
- Dropped the semver-checks job header's pin-rationale paragraph, which still blamed 0.50.0/MSRV after the inline comment right below it was corrected to 0.47.0 and the real reason (rustdoc-JSON format, not MSRV — `install-action` fetches a prebuilt binary). Two copies of the same fact were the kind of drift this PR's own review just caught elsewhere, so pointed the header at the inline comment instead of repeating it.

CI is green on `5f9a4929` (`Semver Checks`, `PR Governance`, `Tests`, `CI Success` all passing), `mergeable_state: clean`. Ready for merge whenever you are.

---

cyberlife-coder:
Correction: my previous comment on this PR carries a "Generated by Claude Code" footer, which AGENTS.md rule 5 forbids on every surface, comments included — my mistake, not something the repository's tooling appended. I have no way to edit or delete that comment through the tools available to me here, so flagging it plainly instead: please disregard that line; nothing else in the comment's content changes.

---

Issue #2087, opened 2026-08-21
config: wire the [search]/[hnsw]/[storage]/[quantization] sections as engine defaults — today only [limits] is applied


cyberlife-coder:
> **Status 2026-09-05 — the `[hnsw]` slice shipped; the other four sections are still inert.**
>
> `Database::resolve_hnsw_params` now reads `self.config.hnsw` (`database/vector_ops.rs:58-74`, called from `graph_ops.rs:54,82` and `vector_ops.rs:119,168`), covered by `database/hnsw_config_wiring_tests.rs`. The `VELESDB_*` provider follow-up (#2185) is closed. The validation warning was narrowed to the field (`config_validation.rs:99-125`).
>
> Still dead, and the code says so itself: `inert_engine_entries()` lists `[search]`, `hnsw.max_layers`, `[storage]` and `[quantization]` (`config_validation.rs:131-158`). The only behavioural consumers of `VelesConfig` are `limits` (`database/collection_ops.rs:45,65,93`) and now `hnsw`. `mmap_cache_mb` / `vector_alignment` are still validated (`config_validation.rs:280`) and read nowhere; same for `query_timeout_ms` (`:197`).
>
> So the list below holds except for `hnsw.m` and `hnsw.ef_construction`.

---

From the final wiring audit: `VelesConfig` parses **and validates** four engine sections that no behavior reads.

**Verified dead** (only consumer of `VelesConfig` behavior-side is `config.limits` — `database/collection_ops.rs:45,65,93`, `collection/types.rs`, `search/vector.rs`):

- `search.ef_search`, `search.max_results`, `search.query_timeout_ms` (there is no query timeout anywhere in core)
- `hnsw.m`, `hnsw.ef_construction`, `hnsw.max_layers`
- `storage.data_dir`, `storage.storage_mode`, `storage.mmap_cache_mb`, `storage.vector_alignment`
- `quantization.*` (round-tripped through the Python options, never applied to a collection)

Every one of these has a hand-written rule in `config_validation.rs` — the config **rejects invalid values for knobs that do nothing**, which signals harder than silence that the knob works. `docs/guides/CONFIGURATION.md` documents them at length, including a `VELESDB_*` env-var table. Runtime overrides (`WITH (ef_search = N)`) are unrelated and work — the query engine has its own path.

[...]

Issue #2078, opened 2026-08-21
perf(storage): wire the WalBatcher group-commit front — or retire it with its config


cyberlife-coder:
From the storage perf audit. The per-batch O(N)→O(1) fsync work landed in the perf wave (batch WAL APIs everywhere); what remains is the **cross-caller amortization** layer that was built and never wired:

- `storage/wal_batcher.rs` (`WalBatcher::submit`/`flush`, ~196 lines, tested) has zero call sites outside its own module and tests.
- `WalBatchConfig` + the `[wal_batch]` TOML section are **public API** (parsed, surfaced through CLI/mobile/tauri) configuring behavior that never engages.

Two honest resolutions — pick one:

1. **Wire it**: front `LogPayloadStorage::store` / `MmapStorage::store` (the single-op paths) with the batcher so concurrent writers share one `sync_all`. This changes single-op latency semantics (a submit waits for the group barrier) and needs a concurrency/load test plus doc updates to the durability contract.
2. **Retire it**: delete the module and deprecate the config section (config compat: accept-and-warn for one release, since `[wal_batch]` may exist in user TOML files).

Leaving a documented-but-inert public config indefinitely is the one option that shouldn't survive.

cyberlife-coder:
## Decision: retire — and the evidence changed what "retire" has to mean

Full-repo audit run today against `develop` @ `86b063ea`. Five findings, one of which was not in the issue and is the strongest argument.

### 1. `WalBatcher` is not a group-commit front, and never was

This is the finding that settles it. `submit` (`storage/wal_batcher.rs:124-140`) takes the `Mutex<PendingBatch>` only long enough to `extend_from_slice` and bump a counter, then releases it and returns `Ok`. The submitter whose increment crosses `max_batch_size` is the one that flushes inline; **every other submitter returns `Ok` while its bytes are still in a `Vec<u8>`.**

A group-commit protocol makes the non-leader writers *wait* for the leader's fsync before their submit returns. This one acknowledges writes that a crash would lose. Option 1 in the issue is therefore not "wire the existing module" — it is "build the leader/follower barrier the module never had", on top of wiring.

There is no thread, no condvar, no timer: `parking_lot::Mutex` is the only synchronization primitive in the file. Which leads to:

### 2. `commit_delay_us` is inert *inside the batcher itself*

[...]

Issue #2174, opened 2026-08-31
chore(config): remove `WalBatchConfig` and its `[wal_batch]` table at the next major


cyberlife-coder:
> **Status 2026-09-05 — the major this issue was waiting for has shipped without the removal.**
>
> `v6.0.0` was tagged and released on 2026-09-02, two days after this issue was opened, and it kept the surface: `[workspace.package] version = "6.0.0"` (`Cargo.toml:22`), and `docs/guides/MIGRATION_v6.0.0.md:117-119` explicitly reconfirms that `WalBatchConfig` and `[wal_batch]` stay. The target is therefore **7.0.0** — and this belongs on the release checklist rather than in a doc comment, since a doc comment is exactly what let it slip the first time.
>
> Nothing has been removed. Still present: `pub struct WalBatchConfig` (`config.rs:335`), the `wal_batch` field (`:376`), the `"wal_batch"` entry in `ENGINE_SECTIONS` (`:502`) **and in `ENV_SECTIONS` (`:420`, not mentioned below)**, the accept-and-warn path (`config_validation.rs:164-165`), and the literal test construction at `database/database_tests.rs:1376` (the body says `:1375`).

---

Follow-up to #2078, which retired the `WalBatcher` module (#2154) and deliberately left the config behind. Closing #2078 left this constraint recorded only in a doc comment in `config.rs`; this issue is so it is not rediscovered by someone deleting the field on its own.

## What is left

`wal_batcher.rs` is deleted. What survives is the deprecated config surface, kept so existing TOML files keep loading:

- `WalBatchConfig`
- the `VelesConfig::wal_batch` field
- the `"wal_batch"` entry in `ENGINE_SECTIONS` (`config.rs:413`)

`[wal_batch]` still parses, and `VelesConfig::validate` warns when `enabled = true` so a config file cannot promise behaviour the engine does not deliver. That accept-and-warn is the deprecation cycle #2078 prescribed, and it is already running.

## The constraint that makes this delicate

**The field and the whitelist entry must go in the same change.** `VelesConfig` carries `#[serde(default)]`, so deleting the field while leaving `"wal_batch"` in `ENGINE_SECTIONS` would make existing TOML files **fail to parse outright** — a config that loads today would stop loading, which is a far worse break than the inert knob this is cleaning up.

[...]

PR #2220, opened 2026-09-07
chore(config): deprecate storage.{data_dir,mmap_cache_mb,vector_alignment}

cyberlife-coder:

`chore/*` → targets `develop`. ✅

## Description

Issue #2087's per-knob audit split `[storage]` in two: `data_dir`, `mmap_cache_mb` and `vector_alignment` have no engine counterpart to wire them to at all (`data_dir` also conflicts irreducibly with the path passed to `Database::open`), unlike `storage_mode`, which is still pending its own wiring decision.

This gives the three the same accept-and-warn deprecation cycle `[wal_batch]` (#2078) is already running: `VelesConfig::validate` now warns, naming exactly which of the three deviates from its default, instead of folding them into the generic "still inert" report. `storage_mode` keeps being reported by `inert_engine_entries` as a single field — the same narrowing `hnsw.max_layers` already got.

**No behavior change.** A config file setting any of these three today keeps loading and validating exactly as before; only the warning text and the `inert_engine_entries()`/new `deprecated_storage_entries()` reporting change.

Related: #2087 (leaves this as one of its follow-ups), precedent: #2078/#2082 (`[wal_batch]`'s own accept-and-warn cycle).

## Type of Change

- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation update
- [x] 🔧 Refactoring (no functional changes)
- [ ] ⚡ Performance improvement

## How Has This Been Tested?

- [x] Unit tests — `cargo test -p velesdb-core --features persistence storage -- --test-threads=1` (284 lib tests + integration suites, 0 failed), including two new/updated tests asserting `inert_engine_entries()` / `deprecated_storage_entries()` split correctly.
- [ ] Integration tests
- [ ] Manual testing

**Test Configuration:**
- OS: Linux (container)

Also ran: `cargo clippy -p velesdb-core --all-targets --features persistence,gpu,update-check -- -D warnings -D clippy::pedantic` (clean), `cargo clippy -p velesdb-core --lib --bins --features persistence,gpu,update-check -- -A warnings -D clippy::undocumented_unsafe_blocks` (clean), `PYO3_PYTHON=python3 cargo clippy -p velesdb-python --lib -- -D warnings` (clean), `cargo test --doc --package velesdb-core` (50 passed), `RUSTDOCFLAGS="-D warnings" cargo doc -p velesdb-core --no-deps --all-features` (clean — no broken intra-doc links), `cargo fmt --check`, and the guard scripts `check-ai-attribution.py`, `check-feature-claims.py`, `check-promise-contract.py`, `check-version-sync.py`, `check_prod_unwraps.py`, `check-todo-annotations.py`, `check-doc-freshness.py --guard stamp` (all pass).

Not run in this sandbox: the full workspace `cargo clippy --workspace` (fails to build `tauri-plugin-velesdb`'s `gdk-sys` build dependency — missing system GTK libs unrelated to this change), wasm/node-specific gates, and the benchmark/perf-smoke jobs (no behavior or hot-path change).

## Checklist

- [x] I have made corresponding changes to the documentation (`docs/guides/CONFIGURATION.md`, `CHANGELOG.md`, rustdoc)

## Unsafe Code Checklist

Skipped — no `unsafe` code added or modified.

## High-Risk Change Checklist

Not applicable — no `hnsw`, `Drop`, `unsafe`, or SIMD dispatch path touched. `storage` is in the filename but this PR only changes config parsing/validation/reporting, not the storage engine itself.

## Additional Notes

Scope is deliberately narrower than all of #2087: `storage_mode` is a distinct, still-open decision (not `quantization::StorageMode`) and is intentionally left as-is, still reported as inert rather than deprecated. `[search]` and `[quantization]` wiring/deprecation are separate follow-ups per #2087's own scoping.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,12 +14,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   three have no engine counterpart to wire them to at all (`data_dir` also
   conflicts irreducibly with the path passed to `Database::open`), unlike
   `storage_mode` and the rest of `[hnsw]`/`[search]`, which are still
-  pending a wiring decision. They now get the same accept-and-warn
-  deprecation cycle `[wal_batch]` (#2078) is already running:
-  `VelesConfig::validate` logs a warning naming exactly which of the three
-  is set away from its default, and removal targets the next major. No
-  behavior changes — a config file setting them today keeps loading exactly
-  as before.
+  pending a wiring decision. They now get the same warn-not-reject
+  deprecation cycle `[wal_batch]` (#2078) is already running — though
+  `[wal_batch]` has no validation at all, while `mmap_cache_mb` keeps its
+  pre-existing hard range check (left as is; loosening it is a separate
+  change). `VelesConfig::validate` logs a warning naming exactly which of
+  the three is set away from its default, and removal targets the next
+  major (tracked by `scripts/check-deferred-removals.py`). No behavior
+  changes — a config file setting them today keeps loading exactly as
+  before.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Deprecated
+
+- **`[storage]` `data_dir`, `mmap_cache_mb` and `vector_alignment` — parsed
+  and validated, never applied.** Issue #2087's per-knob audit found these
+  three have no engine counterpart to wire them to at all (`data_dir` also
+  conflicts irreducibly with the path passed to `Database::open`), unlike
+  `storage_mode` and the rest of `[hnsw]`/`[search]`, which are still
+  pending a wiring decision. They now get the same accept-and-warn
+  deprecation cycle `[wal_batch]` (#2078) is already running:
+  `VelesConfig::validate` logs a warning naming exactly which of the three
+  is set away from its default, and removal targets the next major. No
+  behavior changes — a config file setting them today keeps loading exactly
+  as before.
+
 ### Fixed
 
 - **30 broken rustdoc intra-doc links in `velesdb-core`, and the two lints that
--- a/crates/tauri-plugin-velesdb/examples/production-config/velesdb.toml
+++ b/crates/tauri-plugin-velesdb/examples/production-config/velesdb.toml
@@ -48,13 +48,14 @@ max_layers = 0
 
 [storage]
 # "mmap" or "memory". mmap keeps the resident set small — the right default for
-# a desktop app that may sit in the background for hours.
+# a desktop app that may sit in the background for hours. Reserved — parsed
+# but not yet applied by the engine (core issue #2087).
 storage_mode = "mmap"
 
-# Mmap cache in megabytes. Range: 1..=1048576.
+# mmap_cache_mb and vector_alignment are deprecated (core issue #2087): no
+# engine counterpart exists to wire them to. Parsed only for backward
+# compatibility with existing config files; tuning them has no effect.
 mmap_cache_mb = 1024
-
-# Vector alignment in bytes; 64 matches the cache line on x86_64 and ARM64.
 vector_alignment = 64
 
 [limits]
--- a/crates/velesdb-core/src/collection/search/query/scan_cap_observability_tests.rs
+++ b/crates/velesdb-core/src/collection/search/query/scan_cap_observability_tests.rs
@@ -11,70 +11,7 @@
 #![cfg(all(test, feature = "persistence"))]
 
 use crate::point::Point;
-use crate::test_fixtures::fixtures::{make_point_with_payload, setup_collection};
-use std::sync::{Arc, Mutex};
-
-// ============================================================================
-// Minimal warn-capturing tracing subscriber (no extra dev-dependency).
-// ============================================================================
-
-/// Captures WARN-and-worse events as flat `field=value` strings.
-#[derive(Default)]
-struct WarnSink {
-    events: Mutex<Vec<String>>,
-}
-
-struct FlattenVisitor<'a>(&'a mut String);
-
-impl tracing::field::Visit for FlattenVisitor<'_> {
-    fn record_debug(&mut self, field: &tracing::field::Field, value: &dyn std::fmt::Debug) {
-        use std::fmt::Write;
-        let _ = write!(self.0, "{}={:?} ", field.name(), value);
-    }
-}
-
-struct WarnCapture {
-    sink: Arc<WarnSink>,
-}
-
-impl tracing::Subscriber for WarnCapture {
-    fn enabled(&self, metadata: &tracing::Metadata<'_>) -> bool {
-        // In `tracing`, more severe levels compare LESS (ERROR < WARN < INFO).
-        *metadata.level() <= tracing::Level::WARN
-    }
-
-    fn new_span(&self, _: &tracing::span::Attributes<'_>) -> tracing::span::Id {
-        tracing::span::Id::from_u64(1)
-    }
-
-    fn record(&self, _: &tracing::span::Id, _: &tracing::span::Record<'_>) {}
-
-    fn record_follows_from(&self, _: &tracing::span::Id, _: &tracing::span::Id) {}
-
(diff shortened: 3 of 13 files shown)


cyberlife-coder:
Adversarial review before merge — **BLOCKING**, one item fixed on the branch (`670c954e`), four left for you because they are calls rather than defects.

## Blocking, and now fixed: the promise did not set the alarm

This branch promises removal at the next major **four times** — the CHANGELOG, the `config.rs` rustdoc, `CONFIGURATION.md`, and the runtime warning itself, which tells the user the entries *"will be removed at the next major"*. It touched six files, none of them `scripts/check-deferred-removals.py`.

That is precisely what #2174 records and what #2219 was built to end, reproduced by the first deprecation to follow it. Proven rather than argued — a 7.0.0 tree still carrying all three fields, before:

```
PASSED: workspace major is 7; 0 deferred removal(s) not yet due, none overdue.
exit=0
```

after:

```
FAILED: workspace major is 7 and 1 deferred removal(s) are overdue:
  - the three inert [storage] entries (#2087) was promised for 7.0.0 (#2087),
    and these sites are still here:
      crates/velesdb-core/src/config.rs: pub mmap_cache_mb: usize
exit=1
```

The entry lists the Rust fields, the guide, **and the Python option mirror** — the compiler will not fail when the core fields go, since `velesdb-python`'s options are a separate struct that would keep advertising settings with nowhere to land. A refusal vector ships with it, as `guards.json` requires: without one the entry could later be deleted in silence.

**This is as much a gap in #2219 as in this PR.** That guard catches an overdue removal *that is registered*; it is blind to a promise that never registered. A follow-up worth its own review: a check that finds removal promises in the tree and refuses any without a registry entry. The hard part is false positives — a historical migration note may legitimately carry the phrase without promising anything, which is exactly why `MIGRATION_v6.0.0.md` is excluded from the existing entry.

## Left to you — four calls, not defects

**1. `CONFIGURATION.md` now contradicts itself.** Three zones were updated; three were not:

- L399-409 lists `VELESDB_STORAGE_DATA_DIR` under *"a **reserved** key stays reserved when set this way"*. The same knob is now deprecated at L567 and reserved at L405 — this PR creates a second category and that paragraph only knows the first.
- L186-203 still documents `mmap_cache_mb` as *"Range: 64 - 65536 (64 GB max)"* and `vector_alignment` as *"optimal for most CPUs"*. Functional claims about fields this PR declares inert — and the range contradicts `KNOWN_LIMITATIONS.md:227`, which says 1 TiB.
- L686-691's *"Production - High Performance"* recipe still configures all three.

**2. The shipped example configs teach the dead knobs as tuning.** `crates/velesdb-mobile/examples/velesdb.toml:55-64` says *"1024 is the desktop default and far too generous for a phone"* and *"64 matches the cache line on ARM64"*; `tauri-plugin-velesdb/examples/production-config/velesdb.toml:49-58` likewise. Copying the project's own example now trips the project's own deprecation warning.

**3. The Python surface is untouched, and the recorded reason is a file budget.** `4d22817a` says *"options.rs is already over its 1000-line budget… drop it rather than touch the budget"*. `options.rs:673-696` still documents the four fields as engine settings with defaults, and `to_core()` still writes them. A budget is a reason to split a file, not a reason to leave a deprecation half-announced — and the registry entry above now makes the mirror a removal site regardless.

**4. No test proves the warning is emitted.** Both tests assert the pure function. Delete `self.warn_deprecated_storage_fields();` from `validate()` and the suite stays green, while the log *is* this PR's deliverable. `capture_warns` in `scan_cap_observability_tests.rs:71-77` already exists for this.

## Two smaller ones

- *"the same accept-and-warn cycle as `[wal_batch]`"* is half true: `[wal_batch]` has no validation, while `[storage]` keeps a hard one — `mmap_cache_mb = 0` still fails startup for a knob with no effect (`config_validation.rs:278-284`). Keep it, but then stop citing the precedent; or say why a dead field keeps a hard barrier.
- `hnsw_config_wiring_tests.rs:326` — `test_still_unwired_sections_are_reported_wholesale` now asserts a single field, not a section. The name says the opposite of the test.

## What the review confirmed

The three fields are genuinely inert: `[storage].data_dir` is **not** the server's data dir — that is `[server].data_dir` (`velesdb-server/src/config.rs:97,150,255,270`). Narrowing `[storage]` to `storage_mode` loses nothing: the struct has exactly four fields, three move to the new report, one stays. `cargo-semver-checks` poses no risk here — no `#[deprecated]` attribute is added, and `--release-type minor` disarms the Minor lints anyway. The warning is reachable from every loader, including the Python one.

cyberlife-coder:
All four left-to-me items and both smaller ones addressed in `df6e733` (merges `develop` in too — the branch had fallen behind).

1. **`CONFIGURATION.md` self-contradiction** — the env-var paragraph now calls out `VELESDB_STORAGE_DATA_DIR` as stronger than reserved (deprecated). The raw `[storage]` example replaces the stale functional claims (`mmap_cache_mb`'s wrong "64 GB max" range, `vector_alignment`'s "optimal for most CPUs") with per-field deprecated/reserved notes. Both Usage Examples recipes drop the three dead fields rather than model them as tuning.
2. **Mobile/tauri example configs** — `mmap_cache_mb`/`vector_alignment` comments no longer claim to matter ("far too generous for a phone", "matches the cache line"); replaced with a deprecation note. Values unchanged.
3. **Python surface** — `options.rs` now carries the same deprecated/reserved split as the core doc, appended onto the existing per-field doc lines (not new lines), so the frozen 1123-line budget doesn't move. `to_core()`/`to_core()`-driven `validate()` already emits the warning for Python callers (confirmed: `VelesConfigOptions::to_core` calls `core.validate()`), so no code gap there, only the stale doc.
4. **No test proved the warning fires** — added `test_deprecated_storage_fields_emit_a_warning` (and a negative control, `test_default_config_emits_no_deprecation_warning`) using `capture_warns`. Deleting the `warn_deprecated_storage_fields()` call from `validate()` now fails a test. Moved `capture_warns` out of `scan_cap_observability_tests.rs` into `fixtures_tests.rs` as a shared helper rather than duplicating the ~60-line subscriber; the six existing `scan_cap` tests still pass against the shared version.
5. **"Same as `[wal_batch]`" overclaim** — corrected in `config.rs`, `config_validation.rs`, `CONFIGURATION.md` and `CHANGELOG.md`: only the warn-not-reject shape is shared. `mmap_cache_mb` keeps its pre-existing hard range check; noted as deliberate (loosening it would be a second, separate change) rather than dropping the comparison.
6. **Misleading test name** — `test_still_unwired_sections_are_reported_wholesale` → `test_search_reported_wholesale_storage_mode_reported_by_field`.

Verified before pushing: `cargo fmt --check` (workspace), `cargo clippy -p velesdb-core --all-targets --features persistence,gpu,update-check -D warnings -D pedantic` (clean), `PYO3_PYTHON=python3 cargo clippy -p velesdb-python --lib -D warnings` (clean), `cargo test -p velesdb-core --features persistence` (storage/scan_cap/the new+renamed tests, all green), `check-deferred-removals.py`, `check-file-budgets.py`, `check-ai-attribution.py`, all five `check-doc-freshness.py` guards, `check-doc-contract.sh`, `check-mcp-doc-contract.py`, and the full `python3 -m unittest discover -s scripts/tests` (826 tests, OK).

---

cyberlife-coder:
Re-reviewed after `df6e7337`. **All four review points are addressed**, and the last one properly:

- `CONFIGURATION.md` no longer contradicts itself — the annotated reference block now marks both fields deprecated with the 7.0.0 target and drops the wrong range claim, the env table explains why `VELESDB_STORAGE_DATA_DIR` is *stronger* than reserved, and the `[wal_batch]` asymmetry is stated instead of implied.
- Both shipped example configs now say *"tuning them has no effect"* rather than teaching cache sizes and cache-line alignment as advice.
- The Python option mirror carries the deprecation on the fields themselves.
- `test_deprecated_storage_fields_emit_a_warning` uses `capture_warns` and asserts the warning names the deviating field. Deleting the `warn_deprecated_storage_fields()` call now turns a test red, where before it left the whole file green while the PR's actual deliverable stopped firing.

One nit, not worth holding a green PR for: the *Production - High Performance* recipe still sets `mmap_cache_mb = 128` without an inline note. The reference block, the field table and both example configs all say it is dead, so a reader has three chances to know — but a recipe is what gets copied. Worth folding into the next docs pass rather than another CI cycle here.

Merging.

PR #2227, opened 2026-09-08
refactor(agent): dedup TTL far-end filtering across relation helpers

cyberlife-coder:

`refactor/*` → targets `develop` (this PR does).

## Description

`crates/velesdb-core/src/agent/memory_helpers.rs` had four near-identical
relation accessors — `relations_of`, `incoming_relations_of`, and their
`_bounded` twins — each inlining the same "drop edges whose far endpoint is
TTL-expired" filter loop, differing only in which endpoint (`target()` vs
`source()`) the filter checks. `jscpd` flagged these as duplicate blocks.

Extracted a private `filter_live_far_end(edges, ttl, kind, far_end)` helper
parameterized by a `fn(&GraphEdge) -> u64` endpoint selector
(`GraphEdge::target` or `GraphEdge::source`), and updated all four callers to
use it. No public API, signature, or behavior change — same collection reads,
same filter predicate, same `BoundedRelations` construction.

Added direct unit tests for the new helper (keeps untracked edges, drops an
expired far end on each side, scopes expiry to the given `MemoryKind`)
alongside the existing integration coverage in `semantic_memory_tests`.

## Type of Change

- [x] 🔧 Refactoring (no functional changes)

## How Has This Been Tested?

- [x] Unit tests
- New direct unit tests for `filter_live_far_end` (4 cases), plus existing
  coverage (`agent::semantic_memory_tests`, including
  `test_relations_hide_expired_endpoints`,
  `test_incoming_relations_filters_expired_source`, and the bounded-relations
  cases) which exercises all four callers.
- `cargo clippy -p velesdb-core --lib --tests --features persistence,gpu,update-check -- -D warnings -D clippy::pedantic` — clean
- `cargo test -p velesdb-core --features persistence --lib -- --test-threads=1` — 5486 passed, 0 failed, 16 ignored
- `cargo test --doc --package velesdb-core` — 50 passed
- `cargo fmt --all -- --check` — clean

**Test Configuration:**
- OS: Linux
- Rust version: 1.90.0 (pinned MSRV)

## Checklist

- [ ] I have made corresponding changes to the documentation (none needed — internal, non-public helper)

## Additional Notes

Found via a routine duplication scan (`npx jscpd --min-tokens 50 --format rust crates/`), not tied to any open issue.

### DIFF
--- a/crates/velesdb-core/src/agent/memory_helpers.rs
+++ b/crates/velesdb-core/src/agent/memory_helpers.rs
@@ -1,10 +1,10 @@
 //! Shared helpers for agent memory subsystems (EPIC-010).
 //!
 //! Extracts common patterns used by `SemanticMemory`, `EpisodicMemory`, and
-//! `ProceduralMemory` to avoid code duplication across the three modules.
-//!
-//! These helpers are ready for adoption by memory submodules.
-//! Currently tested directly; callers will migrate in a follow-up.
+//! `ProceduralMemory` to avoid code duplication across the three modules. All
+//! three have adopted these helpers for their relation, TTL, and
+//! serialization plumbing; tests here cover the helpers directly, and the
+//! per-kind `*_tests.rs` suites cover them again through each public API.
 
 use crate::collection::Collection;
 use crate::{Database, DistanceMetric, Point};
@@ -670,6 +670,24 @@ pub(super) fn relate_memory_points(
     Ok(edge_id)
 }
 
+/// Drops edges whose far end — as picked out by `far_end` — is TTL-expired
+/// in the given subsystem.
+///
+/// `far_end` is `GraphEdge::target` for an outgoing edge or `GraphEdge::source`
+/// for an incoming one: the live endpoint is the one already known to the
+/// caller, so the filter always guards the *other* side.
+fn filter_live_far_end(
+    edges: Vec<crate::collection::graph::GraphEdge>,
+    ttl: &super::ttl::MemoryTtl,
+    kind: super::ttl::MemoryKind,
+    far_end: fn(&crate::collection::graph::GraphEdge) -> u64,
+) -> Vec<crate::collection::graph::GraphEdge> {
+    edges
+        .into_iter()
+        .filter(|edge| !ttl.is_expired(kind, far_end(edge)))
+        .collect()
+}
+
 /// Returns the outgoing relation edges of a memory point, filtering out edges
 /// whose target is TTL-expired in the given subsystem.
 ///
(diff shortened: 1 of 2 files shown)


cyberlife-coder:
`CI Success` failed on `f125a12` because the **Python Integrations** job failed: not this PR's.

The failure is in the "Build & install velesdb-python from source" step, before any integration test ran — a `rustup` toolchain sync hit a conflict reinstalling `clippy-preview-x86_64-unknown-linux-gnu` (`detected conflict: 'bin/cargo-clippy'`), which then leaves no working `cargo` on `PATH` for maturin to build against:

```
error: failed to install component: 'clippy-preview-x86_64-unknown-linux-gnu', detected conflict: 'bin/cargo-clippy'
...
Cargo, the Rust package manager, is not installed or is not on PATH.
```

This PR only touches `crates/velesdb-core/src/agent/memory_helpers.rs` and its test module — no toolchain, `Cargo.toml`, or `velesdb-python` changes. Died in runner/toolchain setup, not in test code, so I've queued a re-run of the failed job (`rerun_failed_jobs` on run 34184370374).

---

cyberlife-coder:
Correction to my previous comment: it should not have carried a "Generated by Claude Code" footer — this repo's policy is no AI/assistant attribution anywhere, comments included. No tool is available to me here to edit or delete the prior comment; please disregard that line. The technical content of that comment (the Python Integrations toolchain failure and the queued re-run) stands.

---

cyberlife-coder:
Adversarial review before merge — **no blocking findings**, and the two claims this PR makes about itself were checked rather than taken at face value.

## The refactor is a behavioural no-op, verified site by site

All four call sites collapse to the same predicate they had:

| site | before | after |
|---|---|---|
| `relations_of` | `!ttl.is_expired(kind, edge.target())` | `filter_live_far_end(…, GraphEdge::target)` |
| `relations_of_bounded` | same, on the capped slice | same |
| `incoming_relations_of` | `edge.source()` | `GraphEdge::source` |
| `incoming_relations_of_bounded` | same, on the capped slice | same |

A function pointer replaces a closure over the same method; nothing about ordering, short-circuiting or the expiry comparison moves. `truncated: total > cap` still reads the pre-filter total in both bounded helpers — unchanged, and correctly so.

## Both documentation claims hold

The module header stops saying *"callers will migrate in a follow-up"* and asserts adoption. That is true: 46 helper references in `semantic`, 25 in `episodic`, 24 in `procedural`.

It also claims the per-kind suites cover these paths through each public API — the claim that matters, because helper-only tests would let a caller silently stop calling the filter. Also true:

- `semantic_memory_tests.rs:941` — `test_incoming_relations_filters_expired_source`
- `semantic_memory_tests.rs:1114` — `test_relations_hide_expired_endpoints`

Both go through the public surface, so a call site that dropped the filter turns them red. The new `filter_live_far_end` tests are the unit half; these are the half that makes the deduplication safe.

## One observation, pre-existing and out of scope

In both bounded helpers, `truncated` is derived from the total *before* TTL filtering. A caller can therefore receive fewer edges than `cap` with `truncated: false`, when the gap is expired far ends rather than a cap. That is the behaviour this PR inherited, not one it introduced — naming the filter simply makes it easier to notice. Worth a line in the bounded helpers' docs at some point; not a reason to hold this.

## Why I looked hardest at the endpoint asymmetry

A deduplication that unifies two *almost* identical paths is where the difference nobody wrote down gets erased. Here the difference is real and explicit — outgoing guards `target`, incoming guards `source`, because the near end is already known to the caller — and the extracted helper takes it as a parameter with a doc comment saying exactly that, rather than inferring it. That is the right shape.

PR #2228, opened 2026-09-08
test(persistence): exercise the adopted .vectors arena end to end

cyberlife-coder:

`test/*` → targets `develop`. ✅

## Description

Every assurance that #2215 is safe came from unit tests written alongside the change itself. This covers the sequence a real database performs, and that no test did:

**create → close → reopen** (which maps the durable file as the arena) → **insert past the persisted count** so the mapping has to grow → **save** → **reopen** → **read every vector back**.

That is the path where `File::create` would have truncated the bytes a live mapping still pointed at. The failure mode there is a **SIGBUS rather than an assertion**, so no test could catch it after the fact; what a test *can* do is prove the sequence completes and the data survives.

The fixture goes through a real `Database` on the public API, not a `NativeHnsw` built by hand — it exercises the wiring a user reaches, not the layer the change was written at.

## Two cases

**`reopening_a_collection_does_not_touch_its_vectors_file`** hashes `native_hnsw.vectors` before and after an open, and asserts equality.

Opening a collection must never write to it. That is not tidiness: `velesdb-memory`'s migration resume proves a source store unchanged by hashing exactly these files, and **twenty-one of its tests failed** when adoption first shipped without that guarantee.

**`growing_an_adopted_arena_and_saving_keeps_every_vector`** is the one that bites.

Proven by sabotage rather than asserted: removing the in-place header rewrite from the adopted dump path leaves the persisted count stuck at the pre-growth value, and the test fails naming the gap. Restored afterwards, and `git status` confirms the production file is untouched by this PR.

## Type of Change

- [x] 🧪 Test coverage

## How Has This Been Tested?

```
test reopening_a_collection_does_not_touch_its_vectors_file ... ok
test growing_an_adopted_arena_and_saving_keeps_every_vector ... ok
test result: ok. 2 passed; 0 failed
```

Gates run with the repository's own commands, not convenient ones:

- `cargo clippy --workspace --all-targets --features persistence,gpu,update-check --exclude velesdb-python --exclude velesdb-node -- -D warnings -D clippy::pedantic` — clean
- `python3 scripts/check-drop-tightening.py` — PASSED
- `cargo fmt --check` — clean

Three clippy findings were fixed rather than silenced along the way, including a `u64 → usize` cast made explicit with `try_from` instead of an `allow`.

## Why this exists

`.vectors` v2 and the arena adoption changed the on-disk format and made the durable file the live arena, with one-way compatibility. Both landed with thorough unit tests — all written by the same hand that wrote the change. This is the first coverage of the whole round trip through the public surface, and it is what makes a 6.1.0 release a decision rather than a bet.

## Checklist

### DIFF
--- a/crates/velesdb-core/tests/vectors_format_roundtrip.rs
+++ b/crates/velesdb-core/tests/vectors_format_roundtrip.rs
@@ -1,4 +1,12 @@
-#![allow(clippy::cast_precision_loss)] // ids into f32 coordinates, deliberately
+#![cfg(feature = "persistence")]
+#![allow(clippy::cast_precision_loss)]
+// ids into f32 coordinates, deliberately
+
+// Without `persistence` there is no `Database`, no mapped arena and no
+// `.vectors` at all. Most files in this directory carry the same gate. It is
+// not decoration: `quality-deep.yml` runs `cargo miri test
+// --no-default-features -p velesdb-core` WITHOUT `--lib`, which compiles every
+// test target - so an ungated file breaks a job that only runs on Sundays.
 
 //! End-to-end coverage of `{basename}.vectors` across a real database lifecycle.
 //!
@@ -60,6 +68,18 @@ const ADOPTED: u64 = 64;
 const SMALL: u64 = 5;
 /// Enough new points to force the mapping to grow past its persisted capacity.
 const EXTRA: u64 = 32;
+
+// The size witness below reads "the file is larger than the payload it
+// declares". Capacity doubles on growth, so that holds only while the doubled
+// capacity OVERSHOOTS the final count: at EXTRA == ADOPTED the arena lands on
+// exactly 2x, and the witness would report "copied" on an arena that was in
+// fact adopted - a false negative carrying a confident message. The constraint
+// lives here, where changing a constant trips it, rather than in a sentence
+// nobody re-reads.
+const _: () = assert!(
+    EXTRA < ADOPTED,
+    "the adoption witness needs the doubled capacity to overshoot the final count"
+);
 
 fn make_vector(id: u64) -> Vec<f32> {
     (0..DIM)
@@ -145,6 +165,27 @@ fn payload(path: &Path) -> Vec<Vec<f32>> {
                 .collect()
         })
         .collect()
+}
+
+/// Every disposable arena file under `root`.
+///
+/// `ArenaHome::claim` names them `hnsw-{token}.arena`, so their presence is a
+/// public, on-disk witness that a graph really owns one — no `pub(crate)`
+/// access needed, and no property of the growth strategy inferred.
+fn disposable_arenas(root: &Path) -> Vec<PathBuf> {
+    let mut found = Vec::new();
+    let mut stack = vec![root.to_path_buf()];
+    while let Some(dir) = stack.pop() {
+        for entry in std::fs::read_dir(&dir).expect("test: read dir").flatten() {
+            let path = entry.path();
+            if path.is_dir() {
+                stack.push(path);
+            } else if path.extension().is_some_and(|e| e == "arena") {
+                found.push(path);
+            }
+        }
+    }
+    found
 }
 
 fn digest(path: &Path) -> u64 {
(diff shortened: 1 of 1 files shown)


cyberlife-coder:
Adversarial review before merge — **BLOCKING**, and the finding is against the PR's own premise. Not merging. Verified by running it myself rather than taking the review's word.

## The half this PR is named after cannot fail

`growing_an_adopted_arena_and_saving_keeps_every_vector` asserts, after the header check, only `collection.len()` and five `search()` probes. Both go through the collection API — which is served by an index the gap recovery rebuilds from the WAL on open, not from `.vectors`.

Measured, by deleting the durable file outright and reopening:

```
SONDE: .vectors SUPPRIME
SONDE: len() = 64
SONDE: search top1 = Some(7)
test result: ok. 1 passed
```

**The file can be gone and every API-level assertion still passes.** Of 221 lines, the only load-bearing assertion is an 8-byte header read. The sabotage proof in the PR body is real but narrow: it validates the header rewrite and nothing after it.

The correct model already exists in this repo — `vectors_format_tests.rs:210` compares every stored `f32`, value by value.

## Three more, each fixable

**The docstring claims adoption cannot be observed. It can.** An adopted grow doubles arena capacity, so the file is `4096 + capacity*dim*4` with `capacity > count`, while the copy path writes exactly `4096 + count*dim*4`. Measured: 8192 adopted versus 7168 for the copy path at 64+32 points. Without that assert, raising `MIN_ARENA_CAPACITY` above 64 leaves this file green while exercising nothing — the exact scenario its own docstring warns about.

**`assert!(file.exists(), "closing a collection must never delete its durable vector store")` cannot fail for the reason it names.** The hazard is `ArenaHome::drop`, but `create_vector_collection` yields `StorageMode::Full`, and `arena_dir` is `Some` only for `RaBitQ`/`SQ8` (`native_inner.rs:678-681`) — so no `ArenaHome` is ever constructed on this path. To have teeth it needs `StorageMode::SQ8`.

**It does not cover the side of the floor where the cited regression happened.** The 21 `velesdb-memory` failures came from adoption *below* the capacity floor extending the file on open. At 64 points adoption is legal and extends nothing — remove the floor clause from `graph_io.rs:560` and this test stays green. The sibling loops over `[(FLOOR-1, false), (FLOOR+4, true)]`; this PR abandoned that discipline.

## What is genuinely new here, and worth keeping

Grow-after-adoption is real coverage this repo does not have — no existing test grows a mapped arena and then saves through it. That gap is worth closing. It is simply measured by eight bytes right now while the rest of the file is either redundant with the unit suite or inert.

Rewriting against the file rather than the API, with the adoption witness and the below-floor case, then back for review.

cyberlife-coder:
Second adversarial review, second blocking verdict — and it caught **the same class of mistake in the fix for the first one**. Both fixed in `a7cf0049`, plus the two lower-severity findings.

## The disposable-arena test still never reached its hazard

I moved it to `StorageMode::SQ8` believing the storage mode decides whether an `ArenaHome` exists. It does not: `adopt_durable_file` never consults the mode, and a graph that adopted `.vectors` carries no `ArenaHome` at all — `arena_home.rs` says so in as many words.

Measured: at 64 points under `SQ8` the collection directory holds **no `hnsw-*.arena` whatsoever**; at 5 points it holds one.

So I changed a parameter believing it changed the path, without asserting the premise — the exact fault the previous round was about.

Now the store sits **below the capacity floor**, where adoption is refused and a disposable arena is genuinely built, and **the premise is asserted**:

```
raising the count back to 64  ->  test: no hnsw-*.arena exists, so this
collection never owned a disposable arena and the drop hazard is never reached
```

The witness is a file on disk. `backing_path` is `pub(crate)` and out of reach from an integration test, but `ArenaHome::claim` names its files `hnsw-{token}.arena` — ownership is publicly observable, with nothing inferred from a growth strategy.

## The file did not compile without `persistence`

Confirmed, and worse than the review suggested I'd notice: `ci.yml`'s `--no-default-features` check has no `--all-targets`, but `quality-deep.yml` runs `cargo miri test --no-default-features -p velesdb-core` **without `--lib`**, compiling every test target. This would have broken a job that only runs on Sundays. Gated now, like most files in this directory.

## The adoption witness had an unwritten constraint

It reads "the file is larger than the payload it declares" — which holds only while the doubled capacity **overshoots** the final count. At `EXTRA == ADOPTED` the arena lands on exactly 2x and the witness would report *"the arena was copied"* about an arena that was adopted: a false negative carrying a confident message.

That constraint is now a compile-time assertion sitting next to the constants, where changing one trips it, instead of a sentence nobody re-reads.

## One assertion removed because it was wrong, not because it was inconvenient

The rewrite also asserted the arena file was gone after the drop. It is not — and that is not a defect. `ArenaHome::sweep_stale` exists precisely because *"a crash or a kill skips Drop, so a collection directory can carry arenas from a previous run"*, and the sweep runs at the next open. An arena outliving its owner is an **anticipated state**, so pinning its absence would assert a promise the design does not make.

I nearly reported a bug that is not one. The comment records why, so the next reader does not re-add the assertion.

## What the review confirmed still holds

The assertions bite, which was the whole point of the rewrite. Deleting `.vectors`, flipping one stored `f32`, or freezing the header count each make a **named** test fail with its own message — and the arena stores in insertion order, verified over all 96 vectors, so the value-by-value comparison is sound.

Gates: clippy pedantic clean on the workspace with all targets, drop-tightening at baseline, `--no-default-features --tests` compiles, four tests passing.

PR #2257, opened 2026-09-10
fix(recovery): re-index the mapped vectors a racing save left unlinked

cyberlife-coder:

## Description

Lot P9 of #2246, from the second review of #2253.

A save that races `upsert_bulk`'s async build persists mappings for vectors
the graph has not linked yet. Two windows open it:

- `AsyncIndexBuilder::flush_sync` returns `Ok(0)` while another build holds
  the builder, so a save's drain can do nothing;
- nothing stops an `upsert_bulk` between the drain and the save, and that
  path registers its mappings before any graph insert.

Recovery skipped every mapped id (pass 1), and pass 3 compares arena slots,
which the direct writer had already filled. After a crash those points stayed
**stored, mapped and unreachable by graph search**. #2253 drains before two
more saves, which closes the window without concurrent writers — not these two.

The fix is at the root, in recovery. A mapping is not proof of graph
membership; a layer-0 neighbour is — every node an insert linked has one once
the graph holds two. Pass 4 re-indexes each mapped id whose node has none (the
entry point of a one-node graph excepted), through the same upsert pass 3 uses.

## How Has This Been Tested?

- [x] `collection::core::recovery_tests` — the new test passes. With pass 4 removed it fails (`299 mapped ids are still unlinked after open`); with every node treated as linked, the saved index's premise fails.
- [x] `collection::core` — 258 pass; `bulk_insert_v2_integration` and `reorder_durability` pass.
- [x] `RUSTFLAGS=-Dwarnings` builds without `persistence` (core, wasm) and strict clippy on core — clean.
- [x] `JOBS=lint,hygiene scripts/local-ci.sh` — Gates replayed: 33 - failed: 0 - tool missing: 2 - runner-only: 2 (of 37 declared).
- [x] Every proof above ran in a cargo target no other worktree shared. The cost of the linkage scan at open follows in a comment, measured in release on an idle machine.

## Type of Change

- [x] 🐛 Bug fix (non-breaking) — crash recovery

## Unsafe Code Checklist

Skipped — no `unsafe` code.

## High-Risk Change Checklist

- [x] Recovery path: a fourth pass at open; it only re-indexes, through the same upsert as pass 3.

Refs #2246.


cyberlife-coder:
Cost of pass 4 at open, measured in release on an idle machine (Apple silicon, the only job running).

What was timed: collecting every mapping and asking `unlinked_nodes` about each mapped slot — exactly what `relink_unlinked_ids` does before it finds anything to re-index. Five runs per size, freshly built graphs (every node linked, so the re-index itself costs nothing here).

| mapped vectors | scan, median | min | max |
|---:|---:|---:|---:|
| 100 000 | 0.61 ms | 0.60 ms | 1.21 ms |
| 1 000 000 | 10.8 ms | 10.8 ms | 13.0 ms |

It reads the mappings and one layer-0 neighbour list per node, never a vector, so the dimension (8 here) does not enter. Linear in the mapped count, as expected: about 11 ms per million vectors on every open — small next to loading a collection of that size, and the reason the pass runs unconditionally rather than behind a count shortcut that this defect defeats.

The measurement was a temporary `#[ignore]` test, not committed.

Issue #2259, opened 2026-09-10
HNSW: a parallel batch insert can leave a node no graph search reaches, at any ef


cyberlife-coder:
Found while checking #2254's change to the `analyze_*` stats tests, in the #2246 review.

**What was measured.** The stats fixture — 900 or 1 100 points on a smooth 4-D curve, ids in curve order, inserted by one `upsert` — was rebuilt 1 200 times (400 alone, 800 under 4× concurrent load), querying point 500's own vector:

| run | n | `search` top-5 | `search_with_ef(…, 4096)` top-5 | exhaustive top-5 |
|---|---|---|---|---|
| 139 | 900 | 501, 499, 502, 498, 503 | 501, 499, 502, 498, 503 | 500, 501, 499, 502, 498 |
| 164 | 1 100 | 502, 503, 504, 505, 506 | 502, 503, 504, 505, 506 | 500, 501, 499, 502, 498 |
| 190 | 1 100 | 498, 497, 496, 495, 494 | 502, 498, 503, 497, 504 | 500, 501, 499, 502, 498 |

3 misses in 1 200 builds (1 in 400 alone, 2 in 800 under load). The exhaustive scan always finds 500, so its mapping and vector are intact. A beam of 4096 over at most 1 100 points still misses it, and in run 139 it returns all four of 500's chain neighbours but not 500. The distance to 500 is 0, so any walk that expands a node linking to 500 returns it first: no expanded node links to it.

**Hypothesis, not yet proven.** `parallel_insert`'s chunked connect can leave a node with no layer-0 in-edge. Nodes connected concurrently may not see each other, and the reverse edges a node does get can be pruned away when a neighbour's list overflows. A curve with ids in curve order is the worst case, since a node's nearest neighbours are exactly its batch-mates. The search's stagnation cutoff (`ef_construction / 2`) is the other candidate and has to be ruled out.

**Why it matters.** A point with no in-edge is stored, mapped, and missing from every graph search whatever the `ef`; only the exhaustive `Perfect` scan returns it.

**Plan (#2246, lot P11).** Measure layer-0 in-degree and reachability after the build to confirm or refute the mechanism. Then fix construction so every node is reachable, and prove it on this fixture (0 misses) with a build-time and recall comparison.

cyberlife-coder:
**Mechanism confirmed, and the scale is larger than the self-query showed.**

[...]

PR #2260, opened 2026-09-10
fix(hnsw): one slot allocator — the arena places, the mapping follows

cyberlife-coder:

## Description

**Lot P10 of #2246.** The id mappings kept a slot counter of their own: they predicted each insert's slot at registration, the graph pushed wherever its arena was, and a reconcile step corrected the mapping afterwards; `upsert_bulk`'s direct writer wrote at the prediction and never reconciled. Under concurrent `upsert_bulk` and `upsert` the two counters diverge: `tests/concurrent_bulk_slots.rs` — one thread bulk-loading, one upserting, then an exhaustive search for every id — lost 630, 611 and 645 of 6 400 ids in three runs on `develop`.

The arena is now the only allocator. Every path places the vector, then maps the id to the slot it got (`ShardedMappings::assign`):

| path | before | after |
|---|---|---|
| single insert (`HnswIndex`, `NativeHnswIndex`) | register a predicted slot, insert, correct the mapping if the graph disagreed | insert, then `assign` under the index read guard |
| batch insert | predict N slots, `parallel_insert`, reconcile | the graph returns the slots in input order, then `assign` each |
| vacuum | rebuild, swap, release the write lock, then re-register ids `0..n` | rebuild, swap and re-map from the returned slots under one write lock |
| delete (`remove`) | drop both entries holding no guard: a concurrent renumber could map a deleted id again, or erase the reverse entry of the id that took its slot | hold the index read guard across both writes; `soft_delete` borrows the graph, so releasing the guard first does not compile |
| direct writer (`upsert_bulk` V2) | write at the predicted slot, never reconcile; with storage disabled, map an id to a slot never written | push under the arena lock and map the pushed slots under the index read guard; with storage disabled, map nothing |

Holding the read guard from placement to mapping keeps `reorder_for_locality` and `vacuum`, which renumber under the write lock, from moving a slot in between. `remove` holds it too, across its two map writes, which fixes the third point of #2262; a delete made while `vacuum` rebuilds, before it takes its write lock, is still lost when it re-maps, as are inserts and upserts made then, and a save can still race a vacuum's swap (the rest of #2262). With nothing predicted there is nothing to reconcile or roll back: the prediction, reconcile and rollback helpers go, with `ShardedMappings`' `register*` / `restore` family and the tombstone counter that only the removed fast path fed. `assign` asserts in debug builds that a slot never passes to a second id, and writes under the DashMap shard lock over the id's entry, so assigns of one id serialize.

**Public API is unchanged.** `NativeHnswBackend::{insert, parallel_insert}` and `NativeHnsw::parallel_insert` keep their signatures; the index each takes was never honoured, and their docs now say so. The crate calls an internal `place_batch`.

`docs/SOUNDNESS.md` "HNSW Slot Allocation" replaces "HNSW Batch Insertion Ordering".

Found on the way and tracked separately: #2259 (a parallel batch insert can leave nodes no graph search reaches), #2262 (vacuum loses writes made while it rebuilds), #2264 (the V2 path places each vector twice).

## How Has This Been Tested?

- [x] `tests/concurrent_bulk_slots.rs` (new): red on `develop` 3/3 (630, 611, 645 of 6 400 ids lost); green on this branch, the last two runs on the final code.
- [x] `racing_assigns_and_removes_of_one_id_keep_both_maps_consistent` (eight threads) and `deletes_racing_a_renumber_stay_deleted`: 20 of 20 runs each. Against an `assign` that releases its shard lock before the reverse write, the first fails 30 of 30; against a `remove` that skips the guard, the second fails 30 of 30.
- [x] Seen failing: a direct writer that predicts its slot from `next_idx` fails the stress test 3/3; without the ownership assertion in `assign`, `assign_refuses_a_slot_another_id_holds` fails; a refusal check run after the writes turns the debug tests red. Releasing the guard before `assign` (E0505) or before `soft_delete` (E0382) does not compile.
- [x] `cargo test -p velesdb-core --lib`: 5478 passed, 0 failed.
- [x] `cargo clippy --workspace --all-targets` with `persistence,update-check` and with `persistence,gpu,update-check`, `-D warnings -D clippy::pedantic`: clean. `cargo fmt --all --check`, `cargo doc -p velesdb-core --all-features` with `-D warnings`, and every guard `ci.yml` runs: clean.
- [x] `RUSTFLAGS=-Dwarnings`: `cargo check -p velesdb-wasm --no-default-features --target wasm32-unknown-unknown` and `-p velesdb-core --no-default-features`: clean.
- [x] Bulk-insert throughput, `develop` before #2256 vs this branch before its merge of #2256, on an idle release build: the standard bulk path is 3.6% slower ([measurement](https://github.com/cyberlife-coder/VelesDB/pull/2260#issuecomment-5625642238)); on the V2 direct-writer path this PR rewrote, no difference shows above its run-to-run noise, about ±5%, 48.3 against 48.6 ms ([measurement](https://github.com/cyberlife-coder/VelesDB/pull/2260#issuecomment-5625989628)). `async_builder_enqueue_drain`, which runs no code this PR changes, is the control. #2269 tracks the cost.

## Type of Change

- [x] 🐛 Bug fix (non-breaking) — public signatures unchanged.

## Unsafe Code Checklist

Skipped — no `unsafe` added or changed.

## High-Risk Change Checklist

- **Concurrency:** every insert path holds the index read guard from placement to mapping, in the inner → mappings order every reader already takes; vacuum's mapping rebuild moved inside its write lock; `remove` holds the read guard across its two map writes, so no renumber runs between them.
- **Rollback:** nothing is registered before the graph accepts a vector, so a refused vector or batch leaves every mapping as it was.

Refs #2246.


cyberlife-coder:
**Bulk-insert throughput, develop against this branch.** The description promised this check before merge.

- **Benchmark:** `bulk_insert_v2_benchmark`, release build, criterion.
- **Sides:** develop `8b3ac24f` (A) against this branch at `90e6fc52` (B). Later commits up to `f140ce47` change tests and docs only. The merge of #2256 then brings its arena changes, and the direct writer's pushes start normalizing cosine vectors, so every figure here predates both.
- **Method:** A and B alternated over three rounds, in one session with nothing else running, on an Apple M5 Pro.

Medians:

| bench | A: develop | B: this branch |
|---|---|---|
| `upsert_bulk_standard` (WAL + synchronous HNSW) | 92.8 / 92.4 / 91.9 ms | 94.5 / 97.0 / 95.5 ms |
| `async_builder_enqueue_drain` | 71.2 / 69.0 / 71.4 µs | 71.2 / 71.8 / 67.2 µs |

The standard bulk path is slower in every round, by 1.9%, 4.9% and 3.9%: 3.6% on the mean of the medians. The async builder shows no consistent direction, +0.0%, +4.1% and −6.0% by round; it runs no code this PR changes.

That was the cost of this fix before the merge of #2256, and #2269 re-measures it on the merged code. Without the fix, under concurrent bulk loads, develop loses 611 to 645 of 6 400 ids. A profile shows no new hotspot at the top of the stack, but the release build strips symbols. #2269 tracks finding where the 3.6% goes, and recovering whatever is not the price of the fix.

cyberlife-coder:
**The path this PR rewrote: `upsert_bulk_v2_wired`** (the V2 direct writer of `upsert_bulk`). This uses the same binaries and method as the measurement above: develop `8b3ac24f` as A, this branch at `90e6fc52` as B, alternated over three rounds in one session. Medians:

| round | A: develop | B: this branch |
|---|---|---|
| 1 | 46.97 ms | 49.14 ms |
| 2 | 49.42 ms | 46.87 ms |
| 3 | 48.43 ms | 49.76 ms |

No difference shows above the noise:

- The means of the medians are 48.3 ms against 48.6 ms (+0.7%).
- The widest intervals, A in round 1 and B in round 2, span about 44 to 50 ms; the other four about 47.6 to 50.6 ms.
- By round, B moved +4.6%, −5.2% and +2.8%.

At that noise a cost the size of the standard path's would not show either way, so this neither locates the 3.6% nor rules out a cost on this path.

`async_builder_enqueue_drain`, from the first measurement, runs no code this PR changes. It is a control, not evidence about this change.

Issue #2261, opened 2026-09-10
MCP tool schema publishes 137 rustdoc-only links in its descriptions


cyberlife-coder:
Found by the adversarial re-review of #2258 (lot P8 of #2246), measured on `develop`.

`docs/reference/mcp-tools.json` — the tool schema MCP clients receive — carries **137 rustdoc-only links** in its `description` strings. schemars copies each schema-bound type's doc comment verbatim, so intra-doc link syntax reaches the wire:

- shortcut references with no definition, e.g. ``[`CompilePolicy::slim_response`]``, ``[`ContextAction::Retrieve`]``, ``[`crate::limits::MAX_MEDIA_BYTES`]`` — a Markdown renderer shows them as literal bracketed text;
- inline links whose target is a Rust path, e.g. ``[`feedback`](crate::MemoryService::feedback)``, ``[`stable id`](super::fragment_id)``, ``[`McpServer`](super::McpServer)``, ``[`MemoryService::recall_where`](crate::service::MemoryService::recall_where)`` — a Markdown renderer shows them as broken links.

They appear across 20-odd tools (`compile_context`, `compile_transcript`, `explain_compilation`, `recall*`, `load_working_context`, `save_working_context`, `suggest_budget`, `entity`, …), inputs and outputs alike.

**Why a transform, not 137 edits.** The doc comments are right for rustdoc; the defect is that the schema publishes rustdoc syntax. Rewriting the comments would lose the rustdoc links and come back with the next doc comment. The fix belongs where the schema is produced: turn ``[`X`](path)`` and ``[`X`]`` into `` `X` `` in every published description, regenerate the snapshot, and add a drift guard that fails on any rustdoc link syntax in `mcp-tools.json`.

Tracked as a lot of #2246.

PR #2270, opened 2026-09-11
fix(server): publish OpenAPI descriptions without rustdoc link syntax

### LINKED ISSUE #2263: REST OpenAPI document publishes rustdoc link syntax in 4 descriptions
## What happens

utoipa copies doc comments into the OpenAPI document the server publishes at `GET /api-docs/openapi.json`, committed as `docs/openapi.json` and `docs/openapi.yaml`. Four descriptions in it carry intra-doc link syntax that only rustdoc resolves:

| Where | Span |
|---|---|
| `GET /collections/{name}/graph/traverse/stream` | ``[`STATS_INTERVAL`]`` |
| `POST /collections/{name}/match` | ``[`super::helpers::http_status_for_error`]`` |
| `POST /collections/{name}/points/stream` | ``[`Point`]`` |
| schema `ActualStatsResponse` | ``[`NodeStatsResponse::estimated`]`` |

A client, or an SDK generated from the document, shows these as literal brackets. One of them also names a private helper path.

## Relation to #2261

#2261 fixes the same defect for velesdb-memory's MCP tool schemas: it rewrites each link to what rustdoc shows, at publication. The REST document needs the same treatment:

- apply the rewrite to what `ApiDoc::openapi()` returns, so the served document and the committed snapshot change together;
- add a guard over `docs/openapi.json` like the one in #2261.

Found by the review of #2261.

cyberlife-coder:

`fix/*` → targets `develop` (this PR does).

## Description

utoipa copies doc comments into the OpenAPI document exactly as they are written. A server built with the `swagger-ui` feature serves it at `GET /api-docs/openapi.json`, and `docs/openapi.{json,yaml}` holds a committed copy. Only rustdoc resolves an intra-doc link, so four descriptions showed clients literal brackets, one of them around the path of a private helper.

The four are rewritten in plain prose that states what a client can rely on:

| Where | Before | After |
|---|---|---|
| `GET /collections/{name}/graph/traverse/stream` | ``stats: Periodic statistics (every [`STATS_INTERVAL`] nodes)`` | the traversal runs to completion first, then its events are sent; `stats` carries `nodes_visited` and `elapsed_ms`, after every 100th `node` event; `error` is sent alone instead, when the traversal cannot run |
| `POST /collections/{name}/match` | ``other core variants map per [`super::helpers::http_status_for_error`]`` | when the query fails with a core error, the JSON body carries its `VELES-XXX` code, and the status follows its core error variant: `404` for a missing collection, point, edge or node, `409` for a collection or edge that already exists, `400` for invalid input, `503` for a locked database or a guard rail, `500` for anything else. The `503` a guard rail causes is now declared among the responses. |
| `POST /collections/{name}/points/stream` | ``Each line is a JSON-encoded [`Point`].`` | each line is one JSON-encoded point, whose `id` is an unsigned 64-bit integer; a line that is not blank and does not parse as one, a string `id` included, is counted in the response's `malformed` field |
| schema `ActualStatsResponse` (velesdb-core) | ``mirroring [`NodeStatsResponse::estimated`]`` | ``mirroring the `estimated` flag of `NodeStatsResponse` `` |

A few of these choices need a word:
- **The match errors are scoped to a failing query.** What the server's middleware answers before the handler (the JSON extractor, authentication, rate limiting, the body-size limit) is the server's contract, not this endpoint's, so the description does not list it.
- **The match mapping is spelled out.** `http_status_for_error` gives the match endpoint's mapping, but not every endpoint follows it (`/search/batch` answers 400 for any error its batch search returns), so the description can't just point at "the usual" status.
- **`STATS_INTERVAL` is stated as a number.** The constant is private, so the description gives its value. `test_published_description_states_the_stats_interval` reads that number back and fails if the two disagree.
- **The point sentence names no type.** The document has no `Point` schema, and `PointRequest` accepts a string `id`, which a stream line may not use. So the sentence states the `id` rule instead.

### Why fix the doc comments instead of rewriting at publication

The first version of this PR rewrote the links inside `ApiDoc::openapi()`, through a hand-written walk over utoipa's typed tree. Review found three problems:
- It turned a web link with code text, ``[`Point`](https://…)``, into `` `Point`(https://…) ``.
- It rewrote text inside code spans.
- Most walk branches had no test: parameters, request bodies, tags, references, and composite and array schemas.

Fixing it would have taken a Markdown scanner plus a test for every node kind, all for four spans among the 708 descriptions and summaries develop publishes. velesdb-memory's schemas, by contrast, hold 137 code links and bracketed paths in 102 of their 559 descriptions, which is why #2265 (for #2261) rewrites them at publication.

### The guard

`test_openapi_descriptions_carry_no_rustdoc_link` reads every `description` and `summary` of `ApiDoc::openapi()`. It fails on any of:
- a `[` that opens on a code span;
- a bracketed path, even as a web link's text: a label holding `::`, `@`, `#` or `<`, naming a primitive rustdoc links from a sigil (`&`, `&mut`, `&str`, `*const`, `*mut`), or ending in `()`, `!{}` or `!`, once its backticks are dropped and it is trimmed of whitespace and a quote's `>`, as rustdoc reads it;
- a reference-style link (`[x][y]`, `[x][]`) or a reference definition (any `]:`);
- an inline link to anything but a URL or a fragment.

It reads the raw text, so no Markdown construct (a code span, a quote, a list item) can hide one of these forms from it. That has costs:
- a description cannot show one of these forms, even as code (`` `[x](y)` ``, ``[`asc`, `desc`]``, `&[Vec<f32>]`);
- a web link cannot have text holding code or a path's mark, as in `[issue #2261](…)` or `[Try it!](…)`;
- a description cannot write a reference-style link or definition, even to a URL;
- prose that looks like one of these forms fails too (`[0, 1]: …`, `m[i][j]`, `[#2261]`, `[ops@x.dev]`), which a test pins.

A bare `[Point]` passes, since it reads the same as `[sic]`. All 709 published strings pass. velesdb-memory's schema guard in #2265 applies the same rules.

Fixes #2263

## Type of Change

## How Has This Been Tested?

- `cargo test -p velesdb-server --lib`: 175 passed. That includes:
  - the guard over the served document;
  - four tests of the guard itself: each link form is flagged; the prose it documents as a cost is flagged; web links and brackets that are not links are left alone; every `description` and `summary` is read, including inside arrays, under a property named `description`, and behind escaped `~` and `/` keys;
  - the `STATS_INTERVAL` test.
- The openapi-drift job, run locally. The regenerated `docs/openapi.json` differs from develop's in exactly the four descriptions above and the new `503` response, and `openapi.yaml` changes the same five.
- Mutation proof: each of 38 mutations turns a named test red. The counts per row add up to 38.

| Mutation | Count | Red test |
|---|---|---|
| The server doc comment put back to ``[`Point`]``, or the core one to ``[`NodeStatsResponse::estimated`]`` | 2 | `test_openapi_descriptions_carry_no_rustdoc_link` |
| `STATS_INTERVAL` changed to 50 | 1 | `test_published_description_states_the_stats_interval` |
| Each guard rule removed: the code opener and its padding (a quote's `>` included), inline, reference, `]:`, `::`, `@`, `#`, `<`, each sigil primitive, `()`, `!{}`, `!`, the label's trim, its quote `>` or its backticks; or `mailto::X` read as an address | 22 | `test_rustdoc_link_guard_flags_each_link_form` |
| The `<…>` target or its padding left unstripped, one URL prefix dropped (`http://`, `https://`, `mailto:`, `#`), or any `&` or `*` read as a primitive | 7 | `test_rustdoc_link_guard_leaves_web_links_and_brackets` |
| `summary` unread, every string key read, nested values unread, arrays unread, or the `~` or `/` pointer escape dropped | 6 | `test_rustdoc_link_guard_reads_every_description_and_summary` |

- Gates:
  - `cargo fmt --check`;
  - clippy as CI's two strict steps run it, over the test code too: `cargo clippy --workspace --all-targets --features persistence,update-check --exclude velesdb-python --exclude velesdb-node -- -D warnings -D clippy::pedantic`, and again with `gpu` added to the features;
  - `cargo test --workspace --exclude velesdb-python --exclude velesdb-node --lib`, through the commit hook;
  - core rustdoc with `-D warnings`;
  - every Python guard step `ci.yml` runs (14, with its own arguments and file lists; its wasm, Node and perf runners excepted), and the promise-contract and doc-freshness guards from their own workflows.

## Checklist

- [x] I have made corresponding changes to the documentation (`CHANGELOG.md`, regenerated `docs/openapi.{json,yaml}`)

## Unsafe Code Checklist

Skipped: this PR adds no `unsafe` code.

## High-Risk Change Checklist

Skipped: this PR touches no `hnsw`, `storage`, `Drop`, `unsafe` or SIMD dispatch paths.


cyberlife-coder:
The #2246 review now drives this branch: round 1 reworked the fix (fb83849a), and round 2 is going on top of the develop merge pushed at 06:54 (e21ff49e). Please don't push to `fix/openapi-rustdoc-links` while the review and the merge chain run: the chain merges develop itself before merging, and a concurrent push rejects its own.

Issue #2275, opened 2026-09-13
Mobile (UniFFI) SearchQuality::Custom/Adaptive, and mode = 'custom:<ef>', accept an unbounded ef


cyberlife-coder:
Found while fixing #2274 (`ef_search` out-of-range validation).

**What happens**

`#2274` bounds the dedicated `ef_search` option (VelesQL `WITH`, REST,
config, CLI, `velesdb-python`'s `search_with_ef`) to the documented
`[16, 4096]` range. Two related entry points carry the same unbounded-`ef`
shape and were left out of that PR's scope on purpose, to keep it surgical:

1. **`crates/velesdb-mobile/src/types.rs`** — the UniFFI `SearchQuality` enum
   exposed to Swift/Kotlin:
   ```rust
   Custom { ef: u32 },
   Adaptive { min_ef: u32, max_ef: u32 },
   ```
   `From<SearchQuality> for velesdb_core::SearchQuality` casts these straight
   through (`usize::try_from(ef).unwrap_or(usize::MAX)`) with no range check.
   A mobile app can request `Custom { ef: u32::MAX }` and get the same
   uncapped graph traversal `#2274` closes for `ef_search`.

2. **`mode = 'custom:<ef>'` / `'adaptive:<min_ef>:<max_ef>'`** —
   `api_types::parse_advanced_quality` (and `velesdb-python`'s own duplicate
   parser in `crates/velesdb-python/src/collection/search.rs`) parses the
   `<ef>` substring with `.parse::<usize>()` and no range check. Any client
   that can set `mode` — VelesQL `WITH (mode = 'custom:99999999')`, REST
   `"mode": "custom:99999999"`, the CLI's `\set mode custom:99999999` — gets
   the same unbounded traversal through a different spelling of the same
   option, even after `#2274` ships.

**Why it matters**

Same failure mode `#2274` fixes for `ef_search`: an unbounded `ef` value
runs an uncapped HNSW graph traversal, neither exhaustive nor covered by
`limits.max_perfect_mode_vectors`.

**Scope note**

This is *not* simply "reuse `api_types::validate_ef_search`" for `custom:<ef>`
without a decision: `SearchQuality::Custom` is also the value type
`SearchConfig::resolved_quality()` and `ef_to_quality()` produce internally
from an already-validated `ef_search`, and per `docs/VELESQL_SPEC.md` a
"custom" `ef` is documented as the same `[16, 4096]`-bounded knob as
[...]

Issue #2267, opened 2026-09-10
VelesQL and REST silently ignore a search mode they cannot parse


cyberlife-coder:
## What happens

`WITH (mode = 'adaptive')` in VelesQL, or `"mode": "adaptive"` in a REST search body, does not run Adaptive search. The query runs at the collection's default mode, and nothing tells the caller.

- `mode_to_search_quality` (`crates/velesdb-core/src/api_types/mod.rs:120`) accepts:
  - the named modes `fast`, `balanced`, `accurate`, `perfect` and `autotune`;
  - the forms `custom:<ef>` and `adaptive:<min_ef>:<max_ef>`.

  Anything else returns `None`, bare `adaptive` included.
- **VelesQL:** `QueryOptions::from_with_clause` (`crates/velesdb-core/src/collection/search/query/options.rs:44`) reads that `None` as "no quality override".
- **REST:** the server's search pipeline (`crates/velesdb-server/src/handlers/search/pipeline.rs:264` and `:332`) does the same with `.and_then(mode_to_search_quality)`.

So a typo such as `'acurate'` runs at the default quality, with no error. So does the bare `'adaptive'` that `SEARCH_MODES.md` showed in two examples until #2250.

## Root cause

The parser returns an `Option`, and both callers read `None` as "not given" rather than "not understood".

## Direction

1. Make an unparseable mode an error at both entry points: a VelesQL query error and an HTTP 400, each naming the accepted forms. REST clients that send an unknown mode today get a response; this change would reject them, so it belongs in the release notes.
2. Decide whether bare `adaptive` should mean the documented `{ min_ef: 32, max_ef: 512 }`, and parse it if so.
3. Pin both with tests: an unknown mode is refused, and every documented form is accepted.

Found by the twelfth review round of #2250 (#2246).

cyberlife-coder:
The REPL has the matching gap, found by the fourteenth review round of #2250.

- `\set mode adaptive:512:32` passes `parse_parameterized_mode` (`crates/velesdb-cli/src/session.rs:274-285`), which parses both numbers but never checks that `min_ef <= max_ef`.
- The core parser accepts `adaptive:<min>:<max>` only when `min_ef <= max_ef` (`crates/velesdb-core/src/api_types/mod.rs:143`). So the value the REPL stored is dropped when the query runs, and the query runs at the default mode.

The fix for this issue should refuse `min > max` in the REPL too, with the same message the other forms get.

PR #2276, opened 2026-09-13
fix(velesql): bound ef_search on every surface, and let an explicit ef_search win over mode

### LINKED ISSUE #2274: VelesQL WITH (ef_search = …) is never checked: a non-integer is ignored and -1 becomes an uncapped traversal
Found by the adversarial review of #2273 (round 8). It is the silent fallback #2267 fixed for `mode`, on the option that overrides `mode`.

**What happens**
- `WithClause::get_ef_search` (`crates/velesdb-core/src/velesql/ast/with_clause.rs`) reads the option as `get("ef_search").and_then(WithValue::as_integer).map(|v| v as usize)`, under `#[allow(clippy::cast_sign_loss, clippy::cast_possible_truncation)]`. No rule in `velesql/validation*.rs` checks it.
- A value that is not an integer, such as `WITH (ef_search = 'high')`, is ignored. The query runs at its mode, or at the default quality, with no error.
- A negative value parses (`grammar.pest`: `"-"? ~ ASCII_DIGIT+`) and casts to a huge `usize`, so `-1` becomes `usize::MAX`. `ef_to_quality` (`crates/velesdb-core/src/collection/search/vector_filter.rs`) turns it into `SearchQuality::Custom(usize::MAX)`. The CHANGELOG describes that conversion as "an uncapped graph traversal that is neither exhaustive nor covered by `limits.max_perfect_mode_vectors`".
- `docs/VELESQL_SPEC.md` documents `ef_search` as an integer from 16 to 4096, and nothing enforces that range.

**Why it matters**
- A typo runs silently at another quality, as `mode` did before #2267.
- `ef_search = -1` lets a single query traverse the whole graph, past the guard rail `limits.max_perfect_mode_vectors` puts on `perfect`.

**Options**
Refusing input that runs today is breaking, as it is for #2267, so this needs a decision:
1. Check the option in `QueryValidator`, as #2273 does for `mode`. A value that is not an integer, or falls outside the documented range, would fail with its own `V0xx` code.
2. Clamp it to the documented range and log a warning. Every query keeps running.

The fix should also check whether any other entry point (REST, the bindings) takes an unchecked `ef_search`.

cyberlife-coder:

`fix/velesql-ef-search-range-validation` → targets `develop`.

## Description

Fixes #2274.

`WITH (ef_search = ...)` in VelesQL and `"ef_search"` in a REST search body read the option as `get("ef_search").and_then(as_integer)`, cast straight to `usize` under an `#[allow(clippy::cast_sign_loss)]` that hid the actual defect: the grammar accepts a leading `-` on any integer, so `ef_search = -1` cast to `usize::MAX` and ran an uncapped graph traversal, neither exhaustive nor covered by `limits.max_perfect_mode_vectors`. A value that is not an integer (a typo, `ef_search = 'high'`) was silently ignored, the failure mode #2267 fixed for `mode`.

### The bounds

`api_types::validate_ef_search` (input already known to be non-negative: REST, the CLI, the config file) and `api_types::parse_with_ef_search` (VelesQL's signed grammar, and the Python binding's `int`) reject a value outside the documented `[16, 4096]`, and `api_types::ef_search_out_of_range` builds the one message every surface reports. Every entry point checks it:
- **VelesQL.** `validate_select` checks `mode` and `ef_search` through `validate_with_options`: every query shape fails with `V014`, mirroring `V013`, for every value a repeated key gives (the first applies). A value that is not an integer gets the same message, naming it in canonical VelesQL form: `WithValue` now implements `Display`, writing a text the parser reads back as that same value (a string single-quoted with each quote doubled, a float in decimal with a fractional part and never an exponent, an identifier bare only when the grammar reads it back bare, otherwise double-quoted). It is not the query's own spelling: `1.50` shows as `1.5`, `TRUE` as `true`. `get_ef_search`, for callers already past validation, drops a negative value as its documentation says instead of casting it to `usize::MAX`.
- **REST.** `/search` and `/search/ids` answer `400` naming the range before the collection's circuit breaker counts the request, and count one request error; `/search/batch` checks each entry, naming it, though a batch applies no `ef_search`. A value that is not a non-negative integer fails the body's JSON parsing with axum's plain-text `422`. The OpenAPI document publishes the range as the field's `minimum` and `maximum`, and that `422` on the three paths.
- **CLI and config file.** `\set ef_search` and `[search].ef_search` already enforced this range in v6.0.0; they now call the one definition instead of a copy of the bound.
- **Python.** `search_with_ef` reads its argument through `from_py_with`: any `int` outside the range, a negative one or one no `i64` holds (which pyo3 used to reject with `OverflowError` before the check ran), raises the same `ValueError`; one that is not an int raises `TypeError`. The default stays `ef_search=128`.

### The precedence

An explicit `ef_search` now wins over `mode`, or its alias `quality`, wherever a search applies them, as `docs/VELESQL_SPEC.md` has said since its rewrite and as REST already resolved. VelesQL never did: since `search_with_opts` was introduced, a query that set both ran at the `mode` and dropped the `ef_search`. The plain and the filtered vector path now share one resolution, `QuerySearchOptions::resolved_quality`. The CLI's REPL injected its session `mode` into every query naming none, so neither `\set ef_search` nor an inline `WITH (ef_search = N)` ever reached its search; it now adds its session setting only to a query that names none of `mode`, `quality` and `ef_search`, and then the one it holds (its `ef_search` when set, else its `mode`), so no query it rewrites carries both. EXPLAIN and the cost-based feedback already read the explicit `ef_search`; aggregates read neither option.

This holds in every build with `persistence`. The WASM executor reads no `WITH` option, so it neither applies nor checks `ef_search`, as for `mode`.

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [x] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] 📚 Documentation update

Breaking for a client, over REST, VelesQL or the Python binding, that sent an `ef_search` outside `[16, 4096]`: it now gets a `400` (REST), a `422` with `V014` (VelesQL over `/query`) or a `ValueError` (Python), each naming the accepted range. A VelesQL query that sets both `mode` and `ef_search` now runs at the `ef_search`. The CHANGELOG records the bounds as **BREAKING** under `### Changed` and the precedence under `### Fixed`; its note for the next release covers them with #2267.

## How Has This Been Tested?

- [x] Unit tests
- [x] Integration tests

### Round 5

Verification of `0b673956c` found two minors, plus one review note.

1. `WithValue::Float(f64::NAN)` displayed as `NaN`, which the parser reads back silently as `Identifier("NaN")`, a different value, against the rustdoc's "reads back as this same value". This is only reachable by building the value in Rust. Root cause: round 4 treated NaN's display as free text instead of holding it to the same no-other-value rule. Fix, failing closed: NaN displays as `0.0/0.0`, which the grammar refuses. The rustdoc says NaN has no VelesQL form and its display does not parse. Proof: `test_with_value_display_of_nan_does_not_parse` asserts `parse_back(display(NaN))` is `Err`. It is red on the code before the fix: `NaN read back as Ok(Identifier("NaN"))`.
2. The CHANGELOG had no entry for `impl Display for WithValue`, new public API whose output is part of every `V014` message. Fixed: `### Added` now describes the canonical form, the read-back guarantee, and NaN's display that does not parse.
3. Review note: the round-trip tests compared with `PartialEq`, where `-0.0 == 0.0`. `reads_back_as` now compares floats by `to_bits()`, in the table test and the float proptest. Proof: with `Display` mutated to print `-0.0` as `0.0`, both turn red with `Err("0.0 read back as Float(0.0), not Float(-0.0)")`. The same run, holding both mutations, failed 3 tests of 10.

Checks, after merging `origin/develop` (`3624a78b`):
- `cargo fmt --all -- --check`: clean.
- `velesql::with_options_tests`: 10 passed.
- `collection::search::query::with_options_tests`: 57 passed.
- `--test velesql_parser_conformance`: 1 passed.
- CI's strict clippy on `velesdb-core --all-targets`, with and without `gpu`: clean.
- CI's rustdoc step: clean.
- Same deviation as round 4: `update-check` and `bench-sift1m` were left out, because `rustls 0.23.45` is not in the offline cache.
- After those checks, `origin/develop` gained `08b94204e` (#2285, CI toolchain pinning: same `1.90` channel, only comments, doc comments and a README under `crates/`), merged without a conflict. Nothing was re-run for it; the pre-commit hook passed on the round's own commit.

### Round 4

Finding (closing review at `23696dd4`): `Display for WithValue` wrote a normalised form, `1.50` as `1.5`, `TRUE` as `true` and `100000000000000000000.0` as `1e20`, while this body, its rustdoc and `parse_ef_search_value`'s doc said it wrote the value "as the query wrote it". `1e20` cannot even be read back: the grammar has no exponent.

Root cause: the float arm used `{:?}`, which switches to an exponent for large and small values, and nothing checked the rendering against the parser; the docs described an intent no test held.

Fix: `Display` now writes canonical VelesQL that the parser reads back as the same value. A float goes through `f64`'s `Display` (shortest round-tripping decimal, never an exponent), with `.0` added to a whole number so it does not read back as an integer; an infinity renders as a literal no `f64` holds (`f64::MAX` followed by `0.0`), which reads back as that infinity; `NaN`, which the parser never produces, has no form (round 5 makes its display text that does not parse). An identifier is written bare only when `parser::reads_back_as_bare_identifier` says so. That function asks the grammar itself (`Rule::with_value` over the whole text, yielding that identifier), so `TRUE`, `true_x`, `9lives` or `my option` are double-quoted with each `"` doubled, and no hand-written rule guesses. The three texts now say "in canonical VelesQL form".

Proof:
- `test_with_value_display_reads_back_as_the_same_value` (`velesql/with_options_tests.rs`) parses `WITH (opt = <display>)` back for integers (incl. `-1`, `i64::MIN`/`MAX`), floats (`1e20`, `-1e300`, `f64::MAX`, `1e-20`, `5e-324`, `-0.0`, `100.0`, both infinities), booleans, strings with embedded quotes, and identifiers that do and do not read back bare. `test_with_value_display_writes_the_canonical_text` pins `1e20` → `100000000000000000000.0`. Four proptests round-trip any finite float (asserting no `e`/`E`), any `i64`, any string of non-control characters and any non-empty identifier of them (proptest's `\PC`).
- Red first. With the display as it was at `23696dd4`, 4 of the 9 tests fail: `left: "1e20" right: "100000000000000000000.0"`, `shown as 1e20 … expected string or float`, and proptest minimal inputs `v = -4.215288483729069e-292` and identifier `s = "["` (bare `[` does not parse). With the infinity branch removed: `shown as inf.0`, `right: Ok(Float(inf))`. Not proven red: the mutation dropping the `.0` suffix alone. Its run was still queued on the shared CPU lock when the code was restored, so it tested the fixed code. `(Float(100.0), "100.0")` in both tests covers that case.
- Checks at round 4. The pre-commit hook passed: fmt, workspace clippy, undocumented-unsafe clippy, the workspace tests (velesdb-core lib 5546 passed). After merging `origin/develop`: `cargo fmt --all -- --check` clean; `velesql::with_options_tests` 9 passed; `--test velesql_parser_conformance` 1 passed; CI's strict clippy (`-D warnings -D clippy::pedantic`, `RUSTFLAGS=-Dwarnings`) on `velesdb-core --all-targets`, with and without `gpu`, clean; CI's rustdoc step (`RUSTDOCFLAGS=-D warnings`, `--document-private-items`) clean. Deviation: those clippy and rustdoc runs left out `update-check` and `bench-sift1m`. Both need `rustls 0.23.45` (brought by #2321), which is not in the offline cache here. The code this round touches does not depend on either feature.

Round 3's tests, each run red on the code before its fix, then green:
- `test_explicit_ef_search_wins_over_mode_in_the_search` (`with_options_tests.rs`): over a collection capped at `max_perfect_mode_vectors = 1`, `mode = 'perfect'` alone is refused while adding `ef_search = 64` runs, with `mode` or `quality`, end to end on the plain path and through `search_with_filter_and_opts` on the filtered one. Before: `GuardRail("Perfect (brute-force) search rejected: collection has 20 vectors, exceeding max_perfect_mode_vectors cap of 1 …")`.
- `test_session_ef_search_reaches_the_search`, `test_inline_ef_search_beats_the_session_mode`, `test_inline_mode_beats_a_session_ef_search` (`repl_execute_tests.rs`): the same cap shows which setting reaches the search, the rewritten query carries exactly one quality option, and an untouched session still injects `mode = 'balanced'` alone. Before, for instance: `[("mode", String("perfect")), ("ef_search", Integer(512))]`.
- `test_get_ef_search_drops_a_negative_value`. Before: `Some(18446744073709551615)`.
- `test_a_non_integer_ef_search_gets_the_one_range_message`, `'it''s'` parsed back included. Before: `"ef_search must be an integer between 16 and 4096"`, naming no value.
- `test_openapi_ef_search_bounds_are_the_validators` and `test_openapi_paths_taking_an_ef_search_document_422` (`lib_tests.rs`, every path whose body carries an `ef_search`, found in the spec), and `test_a_mistyped_ef_search_gets_the_documented_422` (`api_integration.rs`: `-1`, `"high"` and `1.5` on the three paths get a plain-text 422, and the spec declares one). Before: `minimum` `0`, and `/collections/{name}/search documents no 422`.
- `test_search_with_ef_refuses_an_ef_search_outside_its_range` (Python): `-1`, `15`, `2**63`, `2**64` and `-(2**63) - 1` raise the message `4097` gets. Before: `OverflowError: Python int too large to convert to C long`.

Earlier rounds' tests stand: both validators' tables; `V014` in every query shape and for a value a repeated key shadows; `from_with_clause`'s own refusal; REST's `400` before the circuit breaker, one request error per refusal, a batch entry named, a valid `ef_search` reaching the REST search; the REPL keeping an inline bad value; the config file's range.

Verified locally with CI's own commands, on the committed tree:

| Check | Result |
|---|---|
| `cargo test -p velesdb-core --lib` `velesql` / `collection::search` / `api_types` / `config` | 1229 / 532 / 90 / 147 passed |
| `cargo test -p velesdb-core --features persistence` executor conformance, `perfect_mode_semantics`, `search_config_defaults` | 13 passed |
| `cargo test -p velesdb-cli` | 397 passed |
| `cargo test -p velesdb-server --lib` / `--test api_integration search` / `--test velesql_conformance_tests` | 177 / 25 / 1 passed |
| `cargo test -p velesdb-python --lib`, as CI runs it / `pytest tests/ -m "not slow"` on a `maturin develop` build | 97 passed / 580 passed, 23 skipped |
| the OpenAPI snapshots, regenerated with CI's openapi-drift command | committed, no drift |
| CI's two strict clippy steps (with and without `gpu`), CI's velesdb-core rustdoc step, `cargo fmt --all -- --check` | clean |
| the 14 Python guard steps `ci.yml` runs, with its arguments; `doc-contract.yml`'s 5 freshness guards; `promise-contract.yml` | pass |
| lizard at CCN ≤ 8 and ≤ 8 parameters, the limits Codacy applies | no new warning |

Round 3's mutations, each turning a named test red (14 of 14):

| mutation | turns red |
|---|---|
| `resolved_quality` resolving `mode` first | `test_query_search_options_ef_search_overrides_mode`, `test_explicit_ef_search_wins_over_mode_in_the_search` |
| the plain path, or the filtered one alone, resolving `mode` first (two mutations) | `test_explicit_ef_search_wins_over_mode_in_the_search` |
| `get_ef_search` casting, as before | `test_get_ef_search_drops_a_negative_value` |
| the REPL injecting its mode and its `ef_search` as before | `test_session_ef_search_reaches_the_search`, `test_inline_ef_search_beats_the_session_mode`, `test_inline_mode_beats_a_session_ef_search` |
| the REPL injecting nothing into an untouched session | `test_session_ef_search_reaches_the_search` |
| the schema bounds removed | `test_openapi_ef_search_bounds_are_the_validators` |
| the 422 removed from `/search`, `/search/ids` or `/search/batch` (three mutations) | `test_openapi_paths_taking_an_ef_search_document_422` |
| the non-integer message built apart, as before, or a quote left undoubled (two mutations) | `test_a_non_integer_ef_search_gets_the_one_range_message` |
| `validate_with_options` no longer checking `ef_search` | `test_ef_search_negative_is_rejected`, `test_ef_search_non_integer_is_rejected`, `test_ef_search_shadowed_by_a_repeated_key_is_rejected` |
| Python's overflow left unmapped | `test_search_with_ef_refuses_an_ef_search_outside_its_range` |

Round 2's mutations of the REST checks, the breaker count, the batch, `\set ef_search` and the config file ran on code round 3 did not touch.

`rustc 1.90.0 (1159e78c4 2025-09-14)`, macOS.

## Checklist

## Unsafe Code Checklist

N/A — no `unsafe` code added or modified.

## Additional Notes

Docs updated: `docs/VELESQL_SPEC.md` (the range, `V014`, every value of a repeated key, `ef_search` overriding `mode`), `docs/guides/SEARCH_MODES.md` (a mode applies only with no `ef_search`), `docs/guides/CONFIGURATION.md` (which setting wins, on every surface), `docs/reference/ERROR_CODES.md` (`V014`), `docs/openapi.{json,yaml}` (regenerated), `CHANGELOG.md`, and the Python `.pyi` stub.

Filed separately rather than widening this PR: #2275, the unbounded `ef`, `min_ef` and `max_ef` of `mode = 'custom:<ef>'` and `'adaptive:<min_ef>:<max_ef>'`, and of the mobile (UniFFI) `SearchQuality::Custom` and `Adaptive` variants. Observed while testing, unchanged here: a filtered VelesQL `NEAR` over a small collection takes the planner's graph-first scan, which applies neither `mode` nor `ef_search`; and `/search/text`, `/search/hybrid` and `/search/multi` answer the same undocumented axum `422` to a mistyped field.


cyberlife-coder:
`MCP Doc Contract` failed on `fa61e83` in `test_sync_agent_hooks.HarnessIsWired.test_the_harness_refuses_a_broken_hook` (`scripts/tests/test_sync_agent_hooks.py`), with `returncode 141` (SIGPIPE) instead of the expected `1`. That file has no history on this branch (last touched by an unrelated node/macOS fix, `64953e6`) and this PR/the follow-up commit touch none of `integrations/agent-hooks/` or `scripts/sync-agent-hooks*`, so this isn't this PR's failure — it looks like a broken pipe inside the sandboxed `hooks.test.sh` subprocess the test spawns, not a real regression.

Requested a re-run via the Actions API; it was rejected with "workflow is already running" since the run was still in progress at the time. Watching for it to settle and will re-run once, per the standing policy of at most one re-run to confirm a flake.

PR #2293, opened 2026-09-13
fix(hnsw): keep writes and parameters through a vacuum; consistent saves

### LINKED ISSUE #2262: HNSW vacuum loses writes made while it rebuilds, and resurrects deletes
Found by the adversarial review of #2260 (lot P10 of #2246), by reading `crates/velesdb-core/src/index/hnsw/index/vacuum.rs`; a reproduction test is part of the fix.

`HnswIndex::vacuum` works in three steps:

1. snapshot the live `(id, vector)` pairs under a **read** guard, then release it;
2. rebuild a fresh graph from that snapshot — the long step, with no lock held;
3. under the write lock, swap the graph in, `clear()` the mappings and re-map the snapshot's ids.

Anything that changes the index between steps 1 and 3 is lost to it:

- **an insert or upsert** lands in the old graph and its mapping, both of which step 3 discards. The id stays in storage, so graph and brute-force search miss it until the next open's gap recovery re-indexes it;
- **a delete** removes a mapping step 3 re-creates from the snapshot, so the deleted id comes back into search results until the next open reconciles orphans;
- **`HnswIndex::remove` takes no lock on the graph**, so a delete can also race step 3's `clear` + re-map directly.

It is reachable in production through `POST /collections/{name}/index/rebuild` (`collection/core/flush.rs`), with writers still accepted.

**Direction.** Under step 3's write lock, reconcile the snapshot with the mappings as they stand: keep only ids still mapped to the slot they had at the snapshot, re-insert from the old arena every id mapped or re-mapped since, drop ids deleted since; and make `remove` take the graph lock's read side so it cannot interleave with step 3. Proof: concurrent upserts and deletes during a vacuum, then every id checked through the exhaustive scan and every deleted id absent.

Tracked as a lot of #2246.

cyberlife-coder:
Fixes #2262.

## What was wrong

`HnswIndex::vacuum` snapshots the live vectors, rebuilds a graph from them without the index lock, then swaps it in. `save` writes a graph and its mappings. Seven defects sat on those two paths:

1. **Writes made during a vacuum were undone.** The swap re-mapped the snapshot's ids, so an insert made during the rebuild vanished from every search, an upsert went back to its old vector, and a delete came back.
2. **The tombstone count missed dead slots after a vacuum.** An id deleted during the rebuild leaves a dead slot in the new graph. `tombstone_count` stopped at the highest slot still mapped, so dead slots above it were never counted, and the count is what `needs_vacuum` reads.
3. **A vacuum of an index with no live id did nothing.** It returned before rebuilding, so every dead slot stayed and `needs_vacuum()` stayed true. No in-tree code triggers a vacuum from that count: the two routes that reach a vacuum, `POST /collections/{name}/index/rebuild` and `POST /collections/{name}/vacuum`, neither reads it (this sentence named only the first until round 6). A caller polling `needs_vacuum()`, though, would vacuum forever and reclaim nothing.
4. **`save` could write mappings that disagree with its graph.** It dumped the graph, released the lock, then read the mappings, one map after the other. A write in between saved an id on a slot the saved graph did not hold, or two maps that disagree, and the next load refused the save.
5. **The graph dump could tear.** It wrote the vectors file, released the arena lock, then wrote the graph file. A batch pushed in between was linked into saved nodes, and the load refused the save ("neighbor id out of range").
6. **Two saves into one directory mixed their files.** The sidecars go through `atomic_write`, a uniquely named temp file then a rename, so each of them is whole. But both saves rewrite `.vectors` and `.graph` in place, and both stamp the generation after the one they read, which is the same one. The load then accepted one save's graph beside the other's mappings.
7. **A vacuum reset the index's parameters.** It rebuilt with `HnswParams::auto(dimension)` rather than the index's own M, `ef_construction` and alpha, and a save then persisted those.

## Fix

1. The swap re-maps the ids mapped at that moment. An id written since the snapshot has its vector copied from the old graph, and an id deleted since stays deleted. "Still on its snapshot slot" means "untouched" only while nothing renumbers, so `vacuum` and `reorder_for_locality` now share a private maintenance lock. Writers and saves never take it.
2. The swap sets `next_idx` to the new graph's slot count (`ShardedMappings::clear_for`).
3. With no live id but dead slots, the vacuum rebuilds into an empty graph. It returns early only when the graph holds no slot at all.
4. `persistence::dump_graph` copies the mappings under the read guard the graph is dumped under, before the dump. It reads the forward map once and derives the reverse map from it. The persisted `next_idx` is the slot count the dump wrote. `HnswIndex::save` and `NativeHnswIndex::save` share it.
5. One arena read guard spans both graph files, in the declared order (vectors 10, then layers 20).
6. Saves of one index are serialized by a `saving` lock of their own. It is not the maintenance lock, so a save made during a vacuum's rebuild still saves the old graph.
7. The rebuild reads M, `ef_construction` and alpha from the graph it replaces.

## Proofs, red on the old code, then green

| Test | Red | Green |
|---|---|---|
| `writes_racing_a_vacuum_survive_it` | 300 of 300 inserts lost, 300 of 300 upserts on their old vector, 297 of 300 deletes back in an exhaustive scan | pass |
| `deletes_racing_a_vacuum_keep_the_tombstone_count_exact` | 5 rounds of 5: tombstone count 0 over 1 229 to 1 361 dead slots | pass |
| `vacuum_of_an_index_with_no_live_id_empties_it` | after a vacuum (tombstones, slots, live ids) = (200, 200, 0), expected (0, 0, 0) | pass |
| `writes_racing_a_save_reload_consistent` | 30 of 31 saves refused: 21 mapping cardinality mismatches, 9 mappings on a slot absent from the saved vectors. With `next_idx` read at the copy instead of the dump (a mutant), 31 of 32 saves reloaded one tombstone short | pass, 3 runs of 3 |
| `a_save_racing_a_vacuum_reloads_consistent` | on `develop`'s save: 5 runs of 5, 3 to 6 of some 40 saves per run reloading 997 to 1 000 ids onto other ids' vectors, or refused as corrupt | pass, without the maintenance lock in `save` |
| `batch_inserts_racing_a_save_reload_consistent` | 3 runs of 3: 13 of 173 saves refused, e.g. "neighbor id 5403 out of range for 5400 vectors" | pass, 3 runs of 3 |
| `saves_racing_into_one_directory_reload_consistent` | without the save lock (a mutant): 2 of 20 rounds left a directory the load refused, "layer num_nodes 9152575309965 exceeds file capacity 296548" and "failed to fill whole buffer"; over 3 runs it failed 2, with 10 of 60 rounds torn | pass, 3 runs of 3 |
| `vacuum_keeps_the_index_parameters` | after a vacuum (24, 300, 1.2) for an index built with (12, 90, 1.0) | pass |

The crate has no test-only interleaving point, so the three save races are bounded stress tests, and the red column gives how often each failed. The unsynchronized graph dump (point 5) also tore once, 1 of 48 saves, in a race run on the tree without its fix.

## Writer wait during a save

This was measured on a shared machine: a release build with LTO off, a 100 000 x 128 Euclidean index, five saves each while a thread inserted one vector at a time.

- **Before point 5:** saves of 70–80 ms. The longest insert waited 35–48 ms (38–56 ms on an index built in memory rather than loaded).
- **After:** saves of 69–71 ms. The longest insert waited 37–42 ms, apart from one first save at 222 ms, whose insert waited 165 ms. I cannot explain that outlier.

**Limits:**
- These are indicative numbers, not a benchmark: five trials, a shared machine.
- Only single inserts were timed. By design, a batch now waits for the whole dump.
- The wait is consistent with single inserts already waiting on the layers lock the graph file's writer holds. The measurement agrees with that mechanism but does not prove it.

## Verification

Run on the final tree. Its diff hash, recorded before the run, was unchanged after it.
- Lib test build: 0 warnings.
- The six race tests, 3 runs: 6 passed, 0 failed each time.
- The vacuum, reorder, save and tombstone tests: 88 passed, 0 failed.
- `hnsw`: 609 passed, 0 failed.
- `cargo fmt --all -- --check`: rc 0.
- CI's two strict clippy steps (with and without `gpu`, `RUSTFLAGS=-Dwarnings`): rc 0.
- CI's velesdb-core rustdoc step: rc 0.
- lizard, CCN and parameters no more than 8, counted against `origin/develop`: `new: 0`.

## Not covered

- `NativeHnswIndex::save` has no save lock of its own. No collection uses that index.
- While a save writes its graph files, an insert waits on the arena lock. Because `parking_lot` locks are task-fair, a search arriving after that insert waits too. This is documented in `docs/guides/CONCURRENCY_LOCKING.md`.

## Round 2

Review of b9150f4d.

### [blocker] A public signature changed

`NativeHnswBackend` is a public, unsealed trait, re-exported as `velesdb_core::index::hnsw::native::NativeHnswBackend`. Its `file_dump` method went from `io::Result<()>` to `io::Result<usize>`, and so did the public inherent `NativeHnsw::file_dump`. That breaks any implementation written against 6.0.0. The CHANGELOG filed the PR under Fixed and named no API change.

**Root cause.** The save needed the slot count the dump wrote, and it read that count through the public methods instead of a crate-private path. No gate caught it: CI's `cargo semver-checks` (0.46.0) passed on b9150f4d.

**Decision: keep the public API unchanged.** Only the crate's own save (`persistence::dump_graph`) reads the count, and that save takes the concrete `NativeHnswInner`, never a user's trait implementation. A break would buy nothing, even with a major release next. Since the API does not change, the CHANGELOG needs no Changed or BREAKING entry.

**Fix.**
- `NativeHnswBackend::file_dump` and `NativeHnsw::file_dump` return `io::Result<()>` again.
- The dump body moved to the new crate-private `NativeHnsw::file_dump_counted`, which returns `usize`. `file_dump` delegates to it and discards the count.
- `NativeHnswInner::file_dump` became `pub(crate) file_dump_counted`. Its type lives in a private module and was never reachable from outside.
- `dump_graph` calls `file_dump_counted`.

**Every other `pub` item the PR changes** (diff against the merge base):
- `NativeHnsw::get_max_connections` and `get_ef_construction` were new public methods on a public type, an unannounced API addition. They are now `pub(crate)`, which is all `NativeHnswInner` needs.
- `ShardedMappings::clear` became `pub(crate) clear_for`. The type is in the private `sharded_mappings` module and only reachable through `pub(crate)` fields, so this is not a public change.
- Nothing else public changed.

**Proof, red first.** The new test `file_dump_keeps_its_public_signature` coerces the trait method and the inherent method, both reached through the public path, to `fn(&NativeHnsw<_>, &Path, &str) -> io::Result<()>`.
- Trait method put back to b9150f4d's `io::Result<usize>`: `error[E0308]: mismatched types … expected fn pointer, found fn item {<… as NativeHnswBackend>::file_dump}`.
- Inherent method put back the same way: `error[E0308]: mismatched types … expected fn pointer, found fn item {graph_io::<impl …>::file_dump}`.
- With both restored, the test passes.

**The vacuum race tests still fail without round 1's fix, re-checked after this change.** I reverted three places in `vacuum.rs`: snapshot ids re-mapped whatever is mapped now, `clear_for(0)`, and the early return when no id is live. All three tests fail:
- `writes_racing_a_vacuum_survive_it`: "after 16 vacuums: 300 inserts lost … 300 upserts still on their old vector".
- `deletes_racing_a_vacuum_keep_the_tombstone_count_exact`: tombstone count 0 over 1 068 to 1 301 dead slots, in 5 rounds of 5.
- `vacuum_of_an_index_with_no_live_id_empties_it`: `(200, 200, 0)` where `(0, 0, 0)` was expected.

After restoring, `git diff --quiet HEAD` passed: the tree matched the commit.

### Verification

Run on the tree merged with `origin/develop` at f75c161b:
- `cargo test -p velesdb-core --lib hnsw`: 610 passed, 0 failed (609, plus the new test).
- Pre-commit hook on this round's commit, velesdb-core lib: 5 535 passed, 0 failed.
- CI's strict clippy flags (`RUSTFLAGS=-Dwarnings`, `--all-targets -- -D warnings -D clippy::pedantic`) on `-p velesdb-core`, with `persistence,gpu` and with `persistence` alone: rc 0 both. The workspace-wide sweep was not run locally.
- CI's velesdb-core rustdoc step (`RUSTDOCFLAGS=-D warnings`, `--all-features --document-private-items`): rc 0.
- `cargo fmt --all -- --check`: rc 0.
- Loom and TSan: not run, and this PR does not rely on them.

**Not re-run.** A last merge of `origin/develop` brought in two commits: the rustls bump (Cargo.lock and manifests) and a change to `storage/snapshot.rs`. Neither touches `hnsw`, and none of the checks above were run after that merge.

## Round 3

Verification review of 3e77b2fc0.

### [blocker] A vacuum could hang for ever beside batch searches

**Root cause.** At 3e77b2fc0, `reconcile` copied the writes made during the rebuild while it held `self.inner.write()`. It did so through `carry_over` and `new.parallel_insert`, and `place_batch` runs on rayon from 100 vectors. `search_batch_parallel` runs its searches on the same rayon pool, and every search takes `inner.read()`. Each worker parked on the guard the vacuum held, so the insert never ran and neither the vacuum nor any search finished. The reviewer reproduced it: 19 threads in `lock_shared_slow`, 1 in `LockLatch::wait_and_reset`, CPU time frozen. The rule broken: never wait on a shared pool while holding a lock that pool's work may take.

**Fix** (`index/hnsw/index/vacuum.rs`):
- After the rebuild, `catch_up` copies the writes made since the snapshot into the new graph, before the write guard is taken, round after round. Each round:
  - holds `inner.read()` only while it lists the written ids and copies their vectors out;
  - releases the guard, then inserts them with `parallel_insert` into the vacuum's own graph, holding no index guard;
  - records where each vector went.
- The rounds stop once at most `CATCH_UP_REMAINDER` (64) writes are left, or after `MAX_CATCH_UP_ROUNDS` (4) rounds, so writes that outpace the copy cannot keep a vacuum catching up for ever.
- Under the write guard, `reconcile` inserts what is left one vector at a time with `HnswInner::insert`. Nothing runs on rayon under that guard. If writes outpaced every round, what is left can exceed 64; it is still copied one insert at a time.
- Unchanged semantics: an id still on the slot its vector was read from takes that vector's new slot, whether the snapshot or a catch-up round read it. An id on another slot was upserted and is copied again. A carried id no longer mapped was deleted and stays deleted, and `clear_for(slots)` counts its node as a tombstone.
- The rebuild itself already ran `parallel_insert` with no guard held.

**Docs updated:** `CONCURRENCY_LOCKING.md` (a catch-up row; the swap row), `CONCURRENCY_MODEL.md`, `SOUNDNESS.md`, the `vacuum` rustdoc and the CHANGELOG.

**Tests.**
- `tests/vacuum_beside_batch_search.rs` (new): `a_vacuum_carrying_writes_finishes_beside_batch_searches`.
  - It is its own test binary: it sets rayon's global pool to 2 threads and loops `search_batch_parallel` on 32 queries. It vacuums a 4 000-vector index while a thread inserts until the vacuum ends, and retries, up to 5 vacuums, until one has raced at least 500 writes.
  - Each vacuum reports through a channel read with a 120 s timeout. On timeout the test writes a HANG line to the process's stderr and exits 1, so it can neither hang the suite nor wedge the global pool of another test binary.
  - It then checks that an exhaustive scan returns every written id.
  - Red, on 3e77b2fc0's `vacuum.rs`: `HANG: vacuum 0 did not finish within 120s beside batch searches on a 2-thread rayon pool`, exit status 1, after 135 s. It failed by timeout and did not hang.
  - Green: 3 runs of 3 on the first version of the test, and 1 of 1 on the final one after the clippy split (about 11 s of test time each).
- `writes_racing_a_vacuum_of_an_sq8_index_survive_it` (new, `index_tests.rs`): the existing write race, on an SQ8 index with a trained quantizer, so the vacuum copies the racing writes into a quantized graph.
  - Afterwards it checks that every write survives an exhaustive scan, and that the index is still SQ8 with its quantizer trained.
  - It does not read the codes back: the backend that holds them is private to `native_inner`.
  - The race body is shared with `writes_racing_a_vacuum_survive_it` (`assert_writes_survive_vacuums`).
- **The existing vacuum race tests are still red without the fix.** I reverted this round's fix in `vacuum.rs` in five places:
  - the slot comparison;
  - the copy of the remaining writes;
  - the catch-up exit;
  - `clear_for(slots)` → `clear_for(0)`;
  - the early return when no id is live.

  4 of 4 failed:
  - `writes_racing_a_vacuum_survive_it`: "after 7 vacuums: 300 inserts lost … 300 upserts still on their old vector".
  - `writes_racing_a_vacuum_of_an_sq8_index_survive_it`: the same failure.
  - `deletes_racing_a_vacuum_keep_the_tombstone_count_exact`: tombstone count 0 over 1 224 to 1 286 dead slots, 5 rounds of 5.
  - `vacuum_of_an_index_with_no_live_id_empties_it`: `(200, 200, 0)` against `(0, 0, 0)`.

  `vacuum.rs` was restored to HEAD after each mutation, and I checked it matched HEAD afterwards.

**Other places that enter rayon while holding an index guard** (grep of `par_iter`, `par_chunks`, `rayon::`, `place_parallel`, `place_batch`, `parallel_insert` in `index/hnsw`). None of these takes a write guard; they hold a read guard, which a waiting writer blocks behind:
- `HnswIndex::insert_batch_parallel` (`index/batch.rs`) and `NativeHnswIndex::insert_batch` (`native_index.rs`) hold `inner.read()` across `place_parallel`, which runs on rayon from 100 vectors. `parking_lot`'s `RwLock` makes new readers wait behind a writer that is waiting. So a vacuum swap or `reorder_for_locality` queued behind that read guard parks the rayon workers of concurrent batch searches, and the batch insert can wait on them for ever. This is on `develop` since #2260 (c610b34d4). It is not introduced here, and it is left out of this PR, to be filed.
- `NativeHnswIndex::brute_force_search_parallel` holds `inner.read()` across `par_chunks`. Its comment says no `inner.write()` could deadlock it, but `NativeHnswIndex` has no vacuum or reorder of its own taking that guard; I did not audit it further.
- `HnswIndex::brute_force_search_parallel` and `search_batch_*` release their guard, or take it per query inside the rayon task.

### Verification

Run on 915eae857 and the test-only follow-up 81b6a9f77, on `origin/develop` at f75c161b. A last merge brought in #2285 (08b94204e): workflows, Dockerfiles and comments in `sync.rs`, loom tests and `CONCURRENCY_MODEL.md`, with the toolchain still 1.90. Nothing in `hnsw` changed, and the checks below were not re-run after that merge.
- `cargo test -p velesdb-core --lib hnsw`: 611 passed, 0 failed.
- The race tests after the clippy fixes: 4 passed.
- `cargo test -p velesdb-core --test vacuum_beside_batch_search`: passed every run, 3 on the first version of the test and 1 on the final one.
- Pre-commit hook on 915eae857, velesdb-core lib: 5 538 passed, 0 failed.
- CI's strict clippy flags on `-p velesdb-core --all-targets`, with `persistence,gpu` and with `persistence`: rc 0 both. The first run found `too_many_lines` in the new test and `significant_drop_tightening` in the SQ8 test; both were fixed, then re-run. The workspace-wide sweep was not run locally.
- CI's velesdb-core rustdoc step (`-D warnings`): rc 0.
- `cargo fmt --check`: rc 0.
- lizard (`-C 8 -a 8`) on `vacuum.rs`: no warning. A first draft had `vacuum` at CCN 9; the snapshot-to-map step moved into `catch_up`.
- Loom and TSan: not run.

## Round 4

Verification review of 53ad2672f. It confirmed the deadlock is gone: the round-3 reproduction finishes, and the catch-up stays correct with 4 to 16 writers.

### [major] The docs promised a remainder the code does not bound

**What was wrong.** Four places said the swap copies only "a few dozen" writes under the write lock, so searches and writes wait only for those:
- the CHANGELOG;
- `CONCURRENCY_MODEL.md`;
- the catch-up row of `CONCURRENCY_LOCKING.md`;
- the doc of `CATCH_UP_REMAINDER` in `vacuum.rs`.

**Root cause.** `catch_up` stops once 64 or fewer writes are left, but also after `MAX_CATCH_UP_ROUNDS` (4) rounds, whatever is left then. Writes that outpace the copy leave the swap as many as they like. The reviewer's probe: 100 000 ids of dimension 8 and 16 writers gave rounds of 36 734, 20 366, 11 089 and 5 970 writes, and 3 367 left for the swap. The docs also left out that the write lock covers rebuilding the mapping of every live id and dropping the old graph.

**Fix: docs only; the code is unchanged.** The four places, and the `vacuum` rustdoc, now say:
- the catch-up runs at most four rounds and stops once 64 or fewer writes are left;
- the swap copies what is left one insert at a time: up to 64, or whatever the fourth round leaves when writes outpace the copy, with no bound;
- the write lock also covers rebuilding the mapping of every live id and dropping the old graph.

**Why no code change.** Letting the rounds adapt (continue while the remainder shrinks), or sealing a watermark as `velesdb-memory`'s online migration does (#1796), would bound the remainder. Either one needs its own measurement against the reviewer's probe. I have not run that probe, so this round states the bound truthfully instead.

**Figure guard.** `scripts/check-figure-sources.py` (markdown-it-py 4.2.0 venv), run on the tree merged with `develop`:
- It first reported 2 figures in `CONCURRENCY_LOCKING.md` that name no measurement. Both were this PR's round-1 save timings, "a save took about 70 ms and kept a single insert waiting up to 35 to 48 ms". That paragraph is gone: only a run on a shared machine backed those numbers, and it is recorded in the round-1 body above.
- After that removal: rc 0, "every performance figure is registered in the promise contract".
- The CHANGELOG is out of the guard's scope.

### Found while committing: the race tests' wait was bounded by a spin count

**What happened.** The pre-commit hook's full `velesdb-core` lib run on the merge failed `batch_inserts_racing_a_save_reload_consistent` with "the saving thread stopped making rounds" (5 537 passed, 1 failed).

**Root cause.** The checkpoint helper `a_round_passes` gave up after 10 000 000 yields. A bound on spins is a bound on time, and on a loaded machine a save outlasted it. `two_saves_while_a_batch_links`, which this PR added, used the same bounded spin. `a_round_passes` came from `develop`, where the renumber race test also uses it.

**Fix.** Both waits now go through `seen_before_it_ends(peer, seen)`. It spins until `seen` holds, and returns `false` only once the peer thread (`ScopedJoinHandle::is_finished`) has ended without it. It reads whether the peer ended before it checks `seen`, so a peer's last round is not missed. Nothing else bounds the wait: a peer that ends early is caught as soon as it ends, and a peer blocked for ever hangs the wait, as it already hung the scope that joins it.

**Proof.**
- Red: with the helper made to treat the peer as ended on its first check (a spin bound at its extreme), `batch_inserts_racing_a_save_reload_consistent` fails with "the saving thread stopped making rounds" (0 passed, 1 failed).
- The hook's full lib run on the fix commit passed.

### Verification

Run on 9d156e07d (9e8822c12 merged with `origin/develop` at f9cc19f24):
- `cargo test -p velesdb-core --lib hnsw`: 611 passed, 0 failed.
- `cargo test -p velesdb-core --test vacuum_beside_batch_search`: 1 passed.
- Pre-commit hook on 9e8822c12 and on the merge 9d156e07d, velesdb-core lib: 5 538 passed, 0 failed, both times.
- CI's strict clippy flags on `-p velesdb-core --all-targets`, with `persistence,gpu` and with `persistence`: rc 0 both. The workspace-wide sweep was not run locally.
- velesdb-core rustdoc `-D warnings`: rc 0. `cargo fmt --check`: rc 0.
- Figure guard: rc 0.
- Loom and TSan: not run.
- **Not merged:** `develop` moved to 0b2b37aa3 (#2276) after this round's merge. Merging it conflicts in the CHANGELOG only. The DRY pre-commit hook refuses that merge commit over duplication #2276 brought in (4 files, each identical to `develop`), and the skip it allows for merge commits was refused by this session's permission system. This head is therefore 1 commit behind `develop`.

## Round 5

Verification review of 9d156e07d. It confirmed the test change of round 4 is sound (no new hang path, the `a_round_passes` change in scope, 624 hnsw tests, the 5 race tests 5 runs each, clippy, rustdoc and the figure guard green) and found one major. **This round changes documentation only: no code, no test.**

### [major] Round 4 replaced one false bound with another

**What was wrong.** Round 4 said the swap copies "up to 64, or whatever the fourth round leaves". That is still false.

**Root cause.** `catch_up` reads what is written with `written_since` and then returns; `vacuum` asks for `self.inner.write()` after that. A batch insert already in flight assigns its mappings between those two moments, and `reconcile` copies every id mapped but not yet carried once the guard is granted. So a catch-up that stopped with **nothing** left can still copy a whole batch under the write lock, and neither `MAX_CATCH_UP_ROUNDS` (4) nor `CATCH_UP_REMAINDER` (64) bounds that copy: they say when the catch-up stops trying.

**Proof** (reviewer's probe):
```
PROBE catch_up stopped early with 0 left
PROBE reconcile copies 2000 under the write lock
```
and, on other runs of the same probe, 193, 2 404 and 7 313 ids copied under the guard.

**Fix — documentary.** The texts now say what the code does: the swap copies every id mapped but not yet carried **when the write lock is granted**; the rounds and the threshold say when the catch-up stops trying, not how much is left; a write in flight, a whole batch, joins that set after the last round looked, so the copy has no upper bound. Bounding it — adaptive rounds, or a sealed watermark like the one `velesdb-memory`'s online migration uses (#1796) — is **out of this PR's scope and tracked in #2335**, with an acceptance test on the ids copied under the guard. This PR keeps its own scope: the writes a vacuum used to lose.

Five places changed, in eb7962cec:
1. `CHANGELOG.md`, the #2262 entry;
2. `docs/CONCURRENCY_MODEL.md`, the vacuum write-lock item;
3. `docs/guides/CONCURRENCY_LOCKING.md`, the catch-up row and the swap row;
4. `crates/velesdb-core/src/index/hnsw/index/vacuum.rs`, the docs of `CATCH_UP_REMAINDER` and `MAX_CATCH_UP_ROUNDS`;
5. the same file's `HnswIndex::vacuum` rustdoc.

`docs/SOUNDNESS.md` needed no change: it never claimed a bound.

**Checks on eb7962cec**, all inside one CPU-lock acquisition, so its pre-commit hook ran under the lock too:
- `scripts/check-figure-sources.py` (pinned venv): rc 0;
- `cargo fmt --all -- --check`: rc 0;
- `cargo test -p velesdb-core --lib hnsw`: 611 passed, 0 failed;
- the pre-commit hook's velesdb-core lib leg: 5 559 passed, 0 failed.

### The merge of `origin/develop` (24efb1b4e), commit d108d5486

The CHANGELOG conflict was resolved keeping every entry (this PR's, then #2299, #2274, #1949, #2283, #2263). The only lines of `develop`'s CHANGELOG the merged file drops are the two this PR replaces on purpose: "a delete made while `vacuum` rebuilds, before it takes its write guard, is still lost (#2262)".

The DRY pre-commit hook refuses a merge commit over duplication the merge brings in from `develop`. The skip authorized for that case was used, with its proof. The hook was run first, so that it would name the file itself, and it reported exactly one block:

```
crates/velesdb-memory/src/mcp/context_tools.rs:527-537 ↔ crates/velesdb-memory/src/mcp/context_tools.rs:468-478  (11 lines)
```

```
identical-to-develop crates/velesdb-memory/src/mcp/context_tools.rs
```

The whole-merge proof printed nothing: of the 29 staged files, the 18 that differ from `develop` are all this PR's own (the CHANGELOG, the three concurrency documents, `tests/vacuum_beside_batch_search.rs` and 14 files under `crates/velesdb-core/src/index/hnsw/`). Every other staged file is byte-identical to `develop`.

The commit then ran its pre-commit under the CPU lock, on the whole workspace: 7 suites, 7 719 passed, 0 failed, the velesdb-core lib leg alone 5 571 passed, 0 failed, 17 ignored, in 243 s.

## Round 6

Adversarial review of d108d5486. It confirmed round 5's documentation is true — it read `vacuum.rs` end to end and every `mappings.assign` / `remove` call site — and raised four points. This round changes one integration test, two test helpers, one rustdoc paragraph and the CHANGELOG. **No production behaviour changes.**

### [major] The rayon invariant had no test that dies to it alone

**Finding.** `crates/velesdb-core/src/index/hnsw/index/vacuum.rs`'s "nothing runs on rayon under that write lock" survived its own mutation: replacing `reconcile`'s one-at-a-time `new.insert` with `new.parallel_insert` left `a_vacuum_carrying_writes_finishes_beside_batch_searches` **green**. It only went red with `MAX_CATCH_UP_ROUNDS = 0` applied as well — an accomplice. The CHANGELOG nevertheless cited that test as the proof.

**Root cause — in the test, not in the code.** The test's racing thread wrote **one id at a time**. The catch-up rounds therefore carried almost everything, and the copy the swap makes under the write guard stayed under a hundred vectors — below the threshold at which `native/backend_adapter.rs:155` `place_batch` leaves its sequential branch. A swap copying "on rayon" never reached rayon, and finished like a correct one.

**Fix — the test forces the swap's own copy over that threshold.** The racing writes now arrive as batches of `RACING_BATCH = 256` ids (asserted `>= 100` and a divisor of the racing id range at compile time) through `insert_batch_parallel`, which holds the index read guard from its first placement to its last mapping. The batch in flight when the vacuum asks for `inner.write()` therefore maps its ids *after* the last catch-up round looked and *before* the guard is granted, so the copy made under the guard covers at least one whole batch, whatever the catch-up managed to carry. The writing thread never pauses, so such a batch is nearly always in flight. The batches cycle over `RACING_IDS = 4096` ids above `INDEXED` instead of taking a fresh id each time: the writer runs for the whole vacuum, and an upsert lands in exactly the same copy — the swap carries an id whose slot changed exactly like one it never saw — while the live set stays bounded.

**Proof — one mutation, no accomplice** (`MAX_CATCH_UP_ROUNDS` left at 4 throughout):

Mutant applied to `reconcile`:
```rust
let refs: Vec<&[f32]> = vectors.iter().map(Vec::as_slice).collect();
let slots = new.parallel_insert(&refs).map_err(...)?;
```

Alive on the test as pushed at d108d5486:
```
test a_vacuum_carrying_writes_finishes_beside_batch_searches ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 11.58s
```

Dead on this round's test, same single mutation:
```
HANG: vacuum 0 did not finish within 120s beside batch searches on a 2-thread rayon pool; the vacuums before it finished with these writes racing each: []
error: test failed, to rerun pass `-p velesdb-core --test vacuum_beside_batch_search`
```

Unmutated, this round's test, three consecutive runs: `1 passed; 0 failed` in 12.94 s, 12.77 s and 13.66 s.

The positive control was resized with it. `RACING_WRITES` is now `2 * RACING_BATCH`, two whole batches, so at least one of them was issued from start to finish during the counted vacuum. A first draft asked for a full pass over `RACING_IDS` and failed its own control on the second run — `no vacuum raced 4096 writes in 5 attempts: [2048, 2816, 3584, 3328, 3584]` — one vacuum of this index simply does not last that long. The final scan is now taken over the ids the cursor actually handed out, not over the whole range.

### [major] An exhaustiveness claim nobody enumerated

**Finding.** "Reachable through `POST /collections/{name}/index/rebuild`" (CHANGELOG) and "the only caller is `POST /collections/{name}/index/rebuild`" (this body, point 3) are false.

**Root cause.** Neither was checked against the route table. `crates/velesdb-server/src/routes.rs` registers **two** routes that reach the same vacuum:

```
47:        .route("/collections/{name}/index/rebuild", post(rebuild_index))
79:        .route("/collections/{name}/vacuum", post(vacuum_collection))
```

Both handlers (`handlers/admin.rs:116` and `:165`) call `collection.rebuild_index()` -> `vacuum_hnsw_index()` (`collection/core/flush.rs:477`) -> `HnswIndex::vacuum`.

**Fix.** Both routes are named, in the CHANGELOG entry and in point 3 above. The rest of this PR's prose was then swept for the same class and each claim checked by grep:

| Claim | Check | Verdict |
| --- | --- | --- |
| "the only caller is `.../index/rebuild`" | `grep -n 'post(' routes.rs`, then the handler chain | **false, fixed** |
| "Reachable through `.../index/rebuild`" | same | **false, fixed** |
| "Nothing in the tree vacuums from that count" | `grep -rn needs_vacuum crates sdks integrations` — only its own definition and two rustdoc examples | true, kept |
| "`vacuum` and `reorder_for_locality` share a maintenance lock that writers and saves never take" | `grep -rn 'maintenance\.lock()' crates` — `index/vacuum.rs:207` and `index/mod.rs:240`, nothing else | true, kept |
| "`NativeHnswInner::file_dump` … was never reachable from outside" | `hnsw/mod.rs:13` is `mod native_inner;`, private, no re-export | true, kept |
| "`ShardedMappings::clear_for` … only reachable through `pub(crate)` fields" | `hnsw/mod.rs:16` is `mod sharded_mappings;`; the only non-test use outside it is `native_index.rs:40`, a `pub(crate)` field | true, kept |

### [minor] The PR added one clone

**Finding.** The 13-line index / `insert_batch` / base-assert preamble of `saves_racing_into_one_directory_reload_consistent` duplicated `batch_inserts_racing_a_save_reload_consistent`.

**Fix.** Two helpers in `index_tests.rs`, `insert_wide_batch(index, first, len)` and `wide_index_with_base(base)`, which both race tests now call.

**Measured** with `jscpd 4.2.4 --min-lines 10 --min-tokens 50 --format rust --max-size 5mb --max-lines 200000` over this PR's changed `.rs` files (jscpd's default 100 kB cap skips `index_tests.rs` and hides the clone):

| tree | files | clones | duplicated lines |
| --- | --- | --- | --- |
| `origin/develop` | 13 | 12 | 193 |
| d108d5486 (before) | 14 | **13** | **205** |
| this round | 14 | **12** | **193** |

### [minor] `catch_up`'s rustdoc read as a post-condition

**Finding.** "until at most `CATCH_UP_REMAINDER` are left" is the exact reading `CATCH_UP_REMAINDER`'s own doc warns against; it was the one of the five changed places left without the caveat.

**Fix.** It now reads "until a round sees `CATCH_UP_REMAINDER` or fewer, or `MAX_CATCH_UP_ROUNDS` have run. Neither bounds what is left: a write in flight maps its ids after the round that stopped looked, and `reconcile` copies them all."

### Checks

Run after the last edit of this round, before the merge of `origin/develop` at f970833b7. Each of these ran inside one CPU-lock acquisition each, after the last edit, with `CARGO_TARGET_DIR` private to this worktree and `RUSTFLAGS=-Dwarnings`:

| check | result |
| --- | --- |
| `cargo fmt --all -- --check` | rc 0 |
| `cargo test -p velesdb-core --lib hnsw` | **611 passed, 0 failed**, 4961 filtered out, 136.64 s — the same 611 as this head |
| `cargo test -p velesdb-core --features persistence --test vacuum_beside_batch_search` | 3 consecutive runs, **1 passed, 0 failed** each |
| `cargo clippy -p velesdb-core --all-targets --features persistence,gpu,update-check -- -D warnings -D clippy::pedantic` | rc 0 |
| `RUSTDOCFLAGS=-D warnings cargo doc -p velesdb-core --no-deps --all-features --document-private-items` | rc 0 |
| `scripts/check-figure-sources.py` (pinned venv) | rc 0 |

Loom and TSan: not run.

### Out of scope

None found.

### The merge of `origin/develop` (f970833b7), commit b2aa4da12

Automatic, no conflict: #2319 touched `crates/velesdb-memory/`, `docs/guides/MCP_SERVER_SETUP.md` and that crate's own CHANGELOG, none of which this PR edits. The root `CHANGELOG.md` kept every entry from both sides untouched. The DRY pre-commit hook did **not** refuse this merge commit, so no skip was used. Its pre-commit ran the workspace suites again under the CPU lock.

## Round 7

CI went red on correct code: the 120 s watchdog of
`a_vacuum_carrying_writes_finishes_beside_batch_searches` is a hidden claim
about the machine, and a debug build on a two-vCPU runner does not honour it.
This round makes the bound measured, then pays the debt round 6 left open —
the mutant proof against **this** tree — and corrects two rows of round 6's own
claim table. **No production behaviour changes** (`vacuum.rs` is byte-identical
to b2aa4da12).

### [major] A fixed watchdog was an unstated claim about the runner

**Root cause.** `HANG_BOUND = Duration::from_secs(120)` priced a vacuum of that
index at "seconds", which is true on an idle dev box and false in debug on a
two-vCPU runner. The constant, not the code, decided the verdict.

**Fix.** The test times **one unraced vacuum of the same index**, in the same
process, profile and rayon pool, and holds each raced vacuum to
`HANG_FACTOR * baseline` with `HANG_FACTOR = 40`. The budget now follows the
machine; a parked vacuum never ends on any machine, so a generous bound still
catches it, only later. Both the `println!` and the `HANG:` line quote the
baseline and the derived bound, so a future red build says which of the two
moved. Measured ratios are recorded in `HANG_FACTOR`'s rustdoc.

### The mutant proof, on this tree

**The mutation.** `reconcile` (`index/vacuum.rs:377`) is the last step of a
vacuum and the only one that runs **under `self.inner.write()`**. Its rustdoc
states the invariant in one clause — "its vector is copied from `old` into
`new` here, **one insert at a time: rayon must not run under the write
guard**". The single edit that reintroduces the hazard therefore replaces that
loop with the batch entry point:

```rust
let refs: Vec<&[f32]> = vectors.iter().map(Vec::as_slice).collect();
let slots = new.parallel_insert(&refs).map_err(...)?;
let slots = one_slot_each(slots, refs.len())?;
for ((id, _), slot) in written.into_iter().zip(slots) { placed.push((id, slot)); }
```

**Why it is the one.** `parallel_insert` reaches `place_batch`
(`native/backend_adapter.rs:155`), which stays sequential under 100 vectors and
spawns rayon tasks above it. Those tasks then park behind `inner.read()` held by
the concurrent `search_batch_parallel`, which the vacuum's own write guard
blocks: the vacuum waits on rayon, rayon waits on the vacuum. Nothing else in a
vacuum runs under that guard, so no other single edit reaches it — `catch_up`
already calls `parallel_insert` deliberately and holds no write guard. It is a
genuine mutation and not a candidate change: applying it contradicts the
rustdoc three lines above it.

**1. The mutant dies** — that edit alone, `MAX_CATCH_UP_ROUNDS` left at 4 and
`CATCH_UP_REMAINDER` at 64, no accomplice
(`cargo test -p velesdb-core --test vacuum_beside_batch_search -- --nocapture`):

```
baseline vacuum 2.389764s, hang bound 95.59056s (40x)
test a_vacuum_carrying_writes_finishes_beside_batch_searches has been running for over 60 seconds
HANG: vacuum 0 did not finish within 95.59056s (40x one unraced vacuum of this index) beside batch searches on a 2-thread rayon pool; the vacuums before it finished with these writes racing each: []
error: test failed, to rerun pass `-p velesdb-core --test vacuum_beside_batch_search`
```

It dies **by the derived bound**, not by a constant: the same run printed the
baseline it was derived from.

**2. Unmutated, it passes — and the generous bound still leaves margin.** Four
runs, the last behind 34 busy spinners on an 18-core box:

| run | baseline | derived bound (40x) | raced vacuum 0 | writes racing it | result |
| --- | --- | --- | --- | --- | --- |
| 1 | 2.362257458 s | 94.49029832 s | 6.508538833 s | 2304 | 1 passed, 0 failed, 16.23 s |
| 2 | 2.325722083 s | 93.02888332 s | 6.417023792 s | 2304 | 1 passed, 0 failed, 14.92 s |
| 3 | 2.342444334 s | 93.69777336 s | 6.466729958 s | 2304 | 1 passed, 0 failed, 14.98 s |
| 4 (34 spinners) | 6.846847959 s | 273.87391836 s | 15.872161375 s | 1792 | 1 passed, 0 failed, 42.54 s |

Baseline and raced vacuum move **together** under load (2.34 -> 6.85 s and
6.47 -> 15.87 s), which is what makes a ratio the right budget and a constant
the wrong one: the observed ratio stayed between 2.3 and 2.8, against a factor
of 40. The mutant, whose vacuum does not end at all, is not reached by that
margin — run 1 of the mutant sat at 95.6 s and would have sat there at any
bound.

### [minor] Two rows of round 6's claim table were wrong

Round 6 swept this PR's prose for unchecked "the only X" claims and published a
table. Two of its **checks** were themselves unchecked. Re-run on this head:

```
$ grep -n "mod native_inner;\|mod sharded_mappings;" crates/velesdb-core/src/index/hnsw/mod.rs
32:mod native_inner;
35:mod sharded_mappings;
```

The table cited `hnsw/mod.rs:13` and `:16`. The verdicts (both modules private,
no re-export) are unchanged and still true; the line numbers were stale.

```
$ grep -rn needs_vacuum crates sdks integrations
crates/velesdb-core/src/index/hnsw/index_tests.rs:125:    assert!(!index.needs_vacuum());
crates/velesdb-core/src/index/hnsw/index_tests.rs:148:    assert!(index.needs_vacuum()); // > 20% threshold
crates/velesdb-core/src/index/hnsw/index_tests.rs:168:    assert!(index.needs_vacuum());
crates/velesdb-core/src/index/hnsw/index_tests.rs:178:    assert!(!index.needs_vacuum());
crates/velesdb-core/src/index/hnsw/index_tests.rs:2472:    // Tombstone ratio = 10 / 20 = 0.50 => needs_vacuum (threshold 0.20)
crates/velesdb-core/src/index/hnsw/index_tests.rs:2474:        index.needs_vacuum(),
crates/velesdb-core/src/index/hnsw/index/vacuum.rs:108:    ///     index.needs_vacuum(); // Consider rebuilding
crates/velesdb-core/src/index/hnsw/index/vacuum.rs:140:    pub fn needs_vacuum(&self) -> bool {
crates/velesdb-core/src/index/hnsw/index/vacuum.rs:194:    /// if index.needs_vacuum() {
```

Round 6 described this as "only its own definition and two rustdoc examples".
It is nine lines: the definition, the two rustdoc examples, and **six lines in
`index_tests.rs`** (five calls plus one comment). The **verdict** still holds —
every one of those six only asserts the predicate, none vacuums from it — so
"Nothing in the tree vacuums from that count" stays true; the description of
the check did not match its own output.

### [minor] `CONCURRENCY_LOCKING.md` catch-up row

`docs/guides/CONCURRENCY_LOCKING.md:189` read "until 64 or fewer are left",
the post-condition reading `CATCH_UP_REMAINDER`'s own rustdoc warns against. It
now reads "until a round **sees** 64 or fewer, or four have run", matching the
`catch_up` rustdoc round 6 fixed.

### Checks

Each inside one CPU-lock acquisition, after the last edit, with a
`CARGO_TARGET_DIR` private to this worktree and `CARGO_BUILD_JOBS=2`:

| check | result |
| --- | --- |
| `cargo fmt --all -- --check` | rc 0 |
| `cargo test -p velesdb-core --lib hnsw` | **611 passed, 0 failed**, 4961 filtered out, 143.16 s — the same 611 as round 6 |
| `cargo test -p velesdb-core --test vacuum_beside_batch_search` | 4 runs, **1 passed, 0 failed** each (table above) |
| `cargo clippy -p velesdb-core --all-targets --features persistence,gpu,update-check -- -D warnings -D clippy::pedantic` | rc 0, 0 warnings |
| `RUSTDOCFLAGS=-D warnings cargo doc -p velesdb-core --no-deps --all-features --document-private-items` | rc 0 |

Loom and TSan: not run. `origin/develop` had not moved since b2aa4da12
(`git rev-list --count HEAD..origin/develop` = 0), so no merge was needed and no
hook skip was used.

### Out of scope

None found.

## Round 8 — the surviving mutant on `reorder_for_locality`

`CI Success` was green on 9a9de34d. The one red check, the non-required
`Mutants on the diff (velesdb-core)`, was a real finding this time rather than
its usual 45-minute cut. From the run's artifact:

```
outcomes.json: "total_mutants": 2, "caught": 1, "missed": 1
missed.txt:    crates/velesdb-core/src/index/hnsw/index/mod.rs:240:9:
               replace HnswIndex::reorder_for_locality -> crate::error::Result<()> with Ok(())
```

**Root cause.** This branch adds the `maintenance: Mutex<()>` and documents it
as serializing "what renumbers slots: `vacuum` and `reorder_for_locality`".
The vacuum arm has the race tests; the reorder arm had nothing. The whole body
of `HnswIndex::reorder_for_locality` could be replaced by `Ok(())` and the
suite stayed green, so no test made the pass do anything — and the function is
user-reachable through `POST /collections/{name}/locality/reorder`
(`crates/velesdb-server/src/routes.rs`). `tests/reorder_durability.rs` does
exercise it, but the mutation job runs `-- --lib --`, which never builds it.

**Fix.** Two `--lib` tests in `crates/velesdb-core/src/index/hnsw/index_tests.rs`,
both on a 1 200-vector index. Sizing: the pass declines below 1 000
(`REORDER_THRESHOLD`, private to `native::graph::reorder`) and returns `None`,
which no caller can tell apart from doing nothing — under that gate the mutant
would be *equivalent*, so the fixture clears it by 20 %, and should the
threshold ever rise past it the moved count falls to zero and the test says so
instead of passing on a silent skip.

- `reordering_renumbers_the_slots_and_the_mappings_follow` snapshots the
  id-to-slot map, reorders, and asserts the observable effect: the slots were
  renumbered (measured **1 176 to 1 199 of 1 200** ids changed slot across
  runs — the base goes in through `insert_batch_parallel`, so the graph, and
  with it the BFS order, is not identical run to run; the assertion is on
  "some moved", never on a count), the new numbering is a permutation of exactly
  the slots it started from, no id was lost or invented, and an id's own
  vector still finds that id, which is what proves `remap_indices` followed
  the renumbering rather than the test merely proving something moved.
- `reorder_waits_for_the_maintenance_lock` drives the arm nothing else
  reaches. It times one unheld reorder and gives a reorder started under a
  held `maintenance` fifty times that to escape, so the bound is measured, not
  a constant in the file — same discipline as round 7's `HANG_FACTOR *
  baseline`. Measured on this machine: `unheld reorder 14.566958ms, window
  728.3479ms (50x)`, printed by the test.

`HnswIndex::lock_maintenance` is a `#[cfg(test)]` accessor added for the second
test alone: the field is private to the `index` module and the tests live one
module up, beside the wide fixtures they share with the vacuum races.

**Proof — the mutant dies.** The body was replaced by `Ok(())` by hand and the
two tests run:

```
running 2 tests
test index::hnsw::index_tests::reorder_waits_for_the_maintenance_lock ... FAILED
test index::hnsw::index_tests::reordering_renumbers_the_slots_and_the_mappings_follow ... FAILED

0 of 1200 mapped ids changed slot

panicked at crates/velesdb-core/src/index/hnsw/index_tests.rs:4524:9:
reorder_for_locality renumbered within 416.7µs while the maintenance lock was
held: it does not take the lock its doc claims, and a vacuum's swap can be
renumbered under it

panicked at crates/velesdb-core/src/index/hnsw/index_tests.rs:4449:5:
reorder_for_locality left every one of the 1200 mapped ids on the slot it
already held: it renumbered nothing

test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 5572 filtered out
```

Either test kills it alone. The body was then restored and the same two run
again: `test result: ok. 2 passed; 0 failed`.

### What round 8 does *not* claim

Both tests bound the reorder arm, nothing more. No test in this branch
establishes that a `vacuum` and a `reorder_for_locality` racing each other
produce a consistent index — only that each takes the lock that would make
them serialize. A race test for that pair is not written here.

### Checks replayed

After the last edit, each inside one CPU-lock acquisition, with a
`CARGO_TARGET_DIR` private to this worktree and `CARGO_BUILD_JOBS=2`:

| check | result |
| --- | --- |
| `cargo fmt --all -- --check` | rc 0 |
| `cargo test -p velesdb-core --lib hnsw` | **613 passed, 0 failed**, 4961 filtered out, 129.43 s — 611 of round 7 plus the two added here |
| `cargo test -p velesdb-core --test vacuum_beside_batch_search` | **1 passed, 0 failed**, 15.45 s — round 7's watchdog, its `new.parallel_insert` mutant guard and its unmutated-under-load run are untouched |
| `cargo clippy -p velesdb-core --all-targets --features persistence,gpu,update-check -- -D warnings -D clippy::pedantic` | rc 0, 0 warnings |
| `RUSTDOCFLAGS=-D warnings cargo doc -p velesdb-core --no-deps --all-features --document-private-items` | rc 0 |

`origin/develop` had moved by one commit (#2282, TypeScript SDK only); it is
merged, with the CHANGELOG conflict resolved by keeping both entries. That
merge commit needed the DRY hook skip authorized for merge-concluding commits:
every file the hook named is byte-identical to `origin/develop`
(`wasm-capabilities-conformance.test.ts`, `wasm-facade.test.ts`,
`wasm-backend.test.ts`, `search-backend.test.ts`, `src/backends/search-backend.ts`),
and every staged file differing from develop is one of this PR's own. No Rust
crate is touched by the merge, so the checks above still describe the tip.

### Out of scope

None found.

## Round 9

No code changed this round. The round was asked to replace the watchdog's
derived bound with an absolute one; measuring first refuted the premise it
rested on, and then found a second deadlock, so raising any bound would have
hidden it.

### The ratio model is unfalsifiable, as suspected — measured

`HANG_FACTOR = 40` was calibrated behind busy spinners that ran for the whole
test, which slows the *unraced baseline* too and keeps the ratio small. CI
times the baseline in a quiet window (nothing races it, the searching and
writing threads have not started) and then races the vacuums against the
test's own load on 2 vCPUs, so the ratio measures the CPU share during the
race and nothing about the code. Reproduced on this test binary, debug, 18
cores, with the baseline always timed quiet and CPU burners added only for
the race:

| burners during the race | the test's CPU share | baseline | raced vacuum | ratio |
| --- | --- | --- | --- | --- |
| 0 | ~200 % | 2.38 s | 6.57 s | 2.8x |
| 34 | 92 % | 2.34 s | 15.72 s | 6.7x |
| 140 | 24 % | 2.34 s | 64.30 s | 27.4x |
| 280 | 11 % | 2.39 s | 97.25 s | 40.7x |

Monotone in the share, no ceiling, and every one of those vacuums finished.
The last row already exceeds 40x on correct code: the factor is a constant in
disguise.

### The red CI is not only sizing: the index still deadlocks

Shrinking the test (1 000 indexed, 128-id racing batches over 1 024 ids,
which still crosses `place_batch`'s 100-vector threshold) made the vacuum
short enough for its write-guard request to land inside a racing batch's
rayon phase. It then deadlocks on an **idle** box: 3 hangs in 6 runs, none of
them ending. `sample` on a hung process gives a closed cycle:

- the vacuum thread: `HnswIndex::vacuum` -> `RawRwLock::lock_exclusive_slow`
  -> `wait_for_readers` — the writer bit is set, which makes parking_lot
  block every *new* reader;
- the racing writer: `insert_batch_parallel` holds `self.inner.read()`
  (`crates/velesdb-core/src/index/hnsw/index/batch.rs:76`) across
  `place_parallel` -> `place_batch` -> `connect_batch_chunked` -> rayon
  `LockLatch::wait_and_reset`;
- both rayon workers: `rerank_candidates_simd`
  (`crates/velesdb-core/src/index/hnsw/index/rerank.rs:253`) ->
  `lock_shared_slow`, parked because of the pending writer;
- the searching thread: parked on its own rayon job.

Nobody can move: the vacuum waits for a read guard that is waiting for a pool
whose workers are waiting for the vacuum.

This is the same class this PR fixes — a lock held across a wait on a shared
pool whose work takes that lock — on the **read** side, and it predates the
PR: `batch.rs:76` and `rerank.rs:253` are identical on `develop`. The PR's
invariant ("nothing enters rayon under the write guard") is therefore
necessary but not sufficient; the sufficient one is that **no rayon job
acquires the index lock at all**, which means hoisting the guard out of
`rerank_candidates_simd` and the per-query search into their batch callers.
That is a locking-contract change across the hot search path (26 `inner.read()`
sites in `index/hnsw/index/`), with its own latency trade-off, so it is left
for arbitration rather than smuggled into this round.

What the current sizes do on this box: 12 runs of 12 finish in 5.97 to 7.08 s
with nothing else running, so the narrow window is never hit there; under 120
burners one run was still progressing after 200 s (workers computing, not
parked) — slow, not deadlocked. Both shapes print the same `HANG` line, so
the red run 35327662707 could be either, and no watchdog value tells them
apart.

### Out of scope

`insert_batch_parallel` holding the index read guard across a rayon wait
(`batch.rs:76`), together with the read guards taken inside rayon jobs
(`rerank.rs:206` and `:253`), deadlocks the index whenever a vacuum asks for
the write guard in between. Present on `develop`, reproducible as described.


cyberlife-coder:
#2343 is the cause of this PR's racing test failing, and fixing it there unblocks this branch. Measured rather than argued.

This branch at `e93b320c0`, merged locally with `fix/hnsw-batch-insert-off-the-global-rayon-pool` (#2344), runs `a_vacuum_carrying_writes_finishes_beside_batch_searches` **4 times out of 4 green**:

| run | taken | bound | measured baseline | margin used |
| --- | --- | --- | --- | --- |
| 0 | 77.1 s | 141.0 s | 3.52 s | 55% |
| 1 | 170.7 s | 227.2 s | 5.68 s | 75% |
| 2 | 109.9 s | 205.9 s | 5.15 s | 53% |
| 3 | 96.4 s | 163.8 s | 4.10 s | 59% |

**Why the fix reaches this test.** It pins the *global* rayon pool to 2 threads, and `RACING_BATCH = 256` is above `PARALLEL_BATCH_MIN`, so the racing writer went through `insert_batch_parallel` — which held `inner.read()` and then joined on those same 2 global workers. #2344 moves that place phase onto a dedicated pool, so it no longer competes for the workers a pending `vacuum` writer has parked. That is the exact edge this test was hitting.

**On comparing this box to the runner.** The failing CI line published its own baseline: `within 177.03398448s (40x one unraced vacuum)` — so the runner measured **4.43 s**, which sits inside the local spread of 3.52–5.68 s. The bound is relative and re-measured each run, so it absorbs the machine difference; these numbers are comparable rather than merely local.

**What this does not settle.** One of four runs used 75% of its bound. That is headroom, not comfort — worth watching if this test ever flakes, but resizing it without evidence would only hide whatever caused the flake. Nothing here changes this PR's own content.

`Mutants on the diff` remains red for #2339, which is independent of both.

cyberlife-coder:
Merged `develop` in (`e806788`, no conflicts) to pick up #2344's rayon-pool-under-guard fix.

- **`Tests`** was failing because `vacuum_beside_batch_search::a_vacuum_carrying_writes_finishes_beside_batch_searches` hung under a 2-thread rayon pool — this branch predates #2344 ("a batch insert no longer waits on the global rayon pool under the index guard"), which landed on `develop` since. Verified locally: the test now completes in ~40s (was timing out past its 109s hang bound before the merge). No PR-specific fix needed, just the freshen.
- **`Mutants on the diff (velesdb-core)`** (in `core-review.yml`, not part of the required `CI Success` chain) times out at the 45-minute action limit on this diff — that's the known, tracked infra issue #2339 ("the mutants job cannot finish a real velesdb-core diff in 45 minutes"), not specific to this PR's changes. No fix exists yet to port.

Also ran the full `index::hnsw::*` lib test suite (579 tests) locally post-merge — all green.

cyberlife-coder:
Round closed on the three findings from the adversarial review, at `2c3f0287`. Each was real; each is fixed with a test that was red first.

## 1. The anti-hang guard was justified by a figure that is not true

Its comment said the raced/unraced ratio "stayed under three" and that 40x left room. Re-measured, four runs on an idle machine:

| baseline | raced | ratio |
| --- | --- | --- |
| 5.73 s | 152.95 s | 26.7 |
| 5.18 s | 107.01 s | 20.6 |
| 5.25 s | 125.25 s | 23.9 |
| 6.02 s | 128.58 s | 21.4 |

The first finished **152.9 s into a 229.2 s bound — 1.50x from firing**, and firing means `report_hang` → `std::process::exit(1)`, a slow vacuum reported as a permanent park. A fifth run, with the rest of the suite building beside it, measured 18.05 s against 108.56 s: a ratio of **6.0**, because load slows the baseline more than it slows the catch-up. The ratio is not a property of the code — it moves between 6 and 27 with what else is running, which is the whole argument against any fixed factor.

The guard no longer measures slowness. The deadlock it exists for stops the **writing** thread too: the batch holding the index read guard is the one waiting on a rayon worker that never comes. So it watches the writer's counter and gives up only when neither the vacuum nor the writes have advanced for three windows of one unraced vacuum each. The time bound survives as a backstop at 400x, fifteen times the worst ratio seen.

A guard that exits the process proves nothing in a passing run, so deciding is now separate from acting: `wait_for_the_vacuum` returns `Waited::GaveUp` and the caller reports. `the_guard_gives_up_when_nothing_advances` drives a silent channel and a frozen counter — it must give up — with the control that a counter advancing every window must not.

Not fixed here, and named by the measurement: the raced cost is dominated by the unbounded catch-up, which is #2335.

## 2. The ordering the save fix is documented on was covered by nothing

Confirmed exactly as reported. The fix is a window on the dump's **far side**, and the position matters more than it looks: my first attempt put it immediately before the dump, and the mutation stayed green — a line written "between the two" travels with its position, so swapping them leaves it on the wrong side. Attached after the dump, a write injected there is harmless with the copy first (nobody names its slot) and fatal with the copy second. With the mutation re-applied:

```
mapping id 401 resolves to index 400 absent from the loaded vector store (graph has 400 vector slots)
```

The hook is keyed by the directory being saved into, so it fires only for that test's tempdir and needs no process-wide serialization.

## 3. `NativeHnswIndex::save` did not serialize with itself

Correct, and the doc stated the invariant of saves rather than of one wrapper. Proved through the **mechanism** rather than the tearing, which is timing-dependent — a test sampling misresolved vectors stayed green without the lock. Each save stamps the generation after the one it reads, so two that run one at a time advance the directory by two:

```
without the saving lock:  12 rounds out of 12 advanced the generation by ONE
with it:                  12 out of 12 advance by two
```

`docs/SOUNDNESS.md` and `docs/guides/CONCURRENCY_LOCKING.md` now name both types and say how the ordering is tested.

## Gates

`fmt`, clippy pedantic (no-gpu, with `internal-bench`), 5576 lib tests at `--test-threads=1`, and the integration binary's two tests — all exit 0. The clippy pass #2350 strengthened two hours ago caught four defects in this very change before CI could: a missed `let...else`, a 100-line test against an 80-line budget, and items after a test module.

`Mutants on the diff` still times out at 45 minutes (59 mutants, 105 s build + 981 s baseline). That is #2339, and branch protection requires only `CI Success`, which this job is not part of.

cyberlife-coder:
`Gate Contracts / Repository guard contracts` was failing for real (`f78056f`, fixed): `persistence.rs`'s new `save_window` module (from the save-ordering fix above) sat inline behind `#[cfg(test)]`, which `scripts/check-inline-tests.py` freezes as new debt in production files.

Moved it to `persistence_save_window_tests.rs` via `#[path]`, the same sibling-file convention `persistence_atomicity_tests.rs` already uses right next to it — no behavior change, `save_window::install`/`pause` keep the same `crate::index::hnsw::persistence::save_window::` path `index_tests.rs` calls. Verified locally: `check-inline-tests.py` passes, `fmt`/clippy pedantic clean, `index::hnsw::persistence*` tests and the specific save-window test green.

---

Issue #2298, opened 2026-09-13
fuzz: fuzz_snapshot_parser fuzzes a stale copy of the snapshot parser, and the copy overflows


cyberlife-coder:
## What

`fuzz/fuzz_targets/fuzz_snapshot_parser.rs` defines its own `parse_snapshot_safe`, with its own `SNAPSHOT_MAGIC`, instead of calling velesdb-core. So the fuzzer has never exercised the parser that loads snapshots (`validate_snapshot_header` in `crates/velesdb-core/src/storage/snapshot.rs`). Of the three fuzz targets it is the only one that never names the library, and the fuzz crate already depends on velesdb-core.

The copy has drifted, and it crashes. The quality-deep run dispatched on #2285's branch is the first where the fuzz job gets past `cargo install cargo-fuzz` (see #2292). Its fuzz job (103805948136) stopped with:

```
thread '<unnamed>' panicked at fuzz_targets/fuzz_snapshot_parser.rs:75:25:
attempt to add with overflow
```

Line 75 is `HEADER_SIZE + entry_count.saturating_mul(ENTRY_SIZE) + CRC_SIZE`. The product saturates at `usize::MAX`, and the `+` after it overflows. Crash input, base64: `VlNOUAF8fHx8fHx8fHx8fHx8fHx8fHx8fHx8fHx8fHx8fP//ef8QrLGvAb6t3g==`

## Production is not affected

The real parser checks `entry_count` against `(len - 25) / 16` before any arithmetic (`snapshot.rs:108-118` on develop), so its `21 + entry_count * 16 + 4` cannot overflow. Only the copy crashes.

## Expected

The fuzzer exercises the code that loads snapshots.

## Fix sketch

1. Expose a byte-slice entry to the parser for fuzzing only, for example `#[cfg(fuzzing)] #[doc(hidden)] pub fn parse_snapshot_bytes(data: &[u8]) -> io::Result<...>` (cargo-fuzz builds with `--cfg fuzzing`). Declare the cfg under `[lints.rust] unexpected_cfgs` so `-Dwarnings` builds stay clean.
2. Make the target call it, and delete the copy.
3. Keep the crash input as a seed in the corpus. Run the target for a fixed time in the PR, and show it finds nothing on the real parser.

cyberlife-coder:
[...]

PR #2324, opened 2026-09-17
docs(storage): say what load_snapshot's NotFound covers

cyberlife-coder:
## Description

Follow-up of #2301 (#2298). Its last review round asked `load_snapshot`'s doc to say what `NotFound` covers; the doc that merged says "`NotFound` if there is no snapshot file", which is narrower than the code.

`load_snapshot` returns `NotFound` when `snapshot_path.exists()` is false, and `Path::exists` returns `false` not only for a missing file but for any error while reaching it (for example a directory without search permission): its std documentation says it "returns `false`" when "there was an error". The caller (`log_payload.rs`) treats `NotFound` as a cold start and replays the WAL, so a snapshot it cannot reach is handled as an absent one.

This PR changes documentation only, to state that behaviour. Whether an unreachable snapshot should instead surface its error (`try_exists()?`) is a behaviour change left for a decision; it is not made here.

## Type of Change

- [x] 📚 Documentation update

## How Has This Been Tested?

- Doc-only change in `crates/velesdb-core/src/storage/snapshot.rs`; CI's `cargo doc -p velesdb-core --no-deps --all-features --document-private-items` with `RUSTDOCFLAGS=-D warnings` checks the new `[`Path::exists`]` link.


cyberlife-coder:
Correction to the body: the std quote is paraphrased there. The exact wording of `Path::exists` is: "If you cannot access the metadata of the file, e.g. because of a permission error or broken symbolic links, this will return `false`." The same page recommends `try_exists()` instead. The squash message will quote it exactly.

PR #2351, opened 2026-09-19
fix(ci): the local gate replayer no longer kills a strict step before it runs

### LINKED ISSUE #2346: scripts/local-ci.sh reports two lint steps FAILED with no captured output, on commands that pass directly and in CI
On macOS (Apple M5 Pro, zsh caller, bash-executed steps), `JOBS=lint,hygiene bash scripts/local-ci.sh` ends with:

```
Gates replayed: 40 - failed: 2 - tool missing: 2 - runner-only: 2 (of 46 declared)
FAILED: Drop cached workspace fingerprints (force full re-lint) Check each velesdb-memory feature in isolation
```

Both steps print **no diagnostic at all** — the harness's `printf '%s\n' "$out" | tail -15` renders an empty line for the first and only `::group::--features mcp` for the second.

## Why this is a harness problem, not a tree problem

Each step passes when executed directly, with the same `bash -c` and the same `CARGO_TARGET_DIR`:

- the clean block (`cargo metadata --locked | jq | xargs -r cargo clean`) returns 0 on a fresh target dir and on an existing one; `xargs -r` is accepted on this machine;
- all eight features of the loop return 0, checked one by one: `mcp http context persistence embedder-http extractor-http ollama extract`.

And CI, which runs the same two steps with every tool present, reports `Lint & Format: SUCCESS` and `Hygiene (typos, taplo, machete): SUCCESS`.

## Why it matters

The script's own header states the principle it is built on: *"A step it cannot replay is ANNOUNCED, never dropped in silence: a gate skipped without saying so is worth less than a gate that is absent, because the operator believes it ran."* A step reported FAILED with no output is the mirror image of that: the operator believes something broke and has nothing to act on. The practical cost is that a real failure in either step becomes indistinguishable from this noise, so the tool's verdict on `lint` stops being usable — which is how a gate gets ignored, then switched off.

It also costs whoever hits it: two hypotheses were formed and refuted here (a shared `CARGO_TARGET_DIR` being wiped by the clean step; `xargs -r` being unsupported on BSD) before the conclusion landed on the harness itself.

## Shape of a fix

`out=$(bash -c "$cmd" 2>&1); rc=$?` (line ~137) captures combined output, so an empty `$out` with a non-zero `rc` means the command produced nothing before dying — worth distinguishing from a command that failed loudly. Reporting the exit code alongside the (empty) output, and keeping the step's stderr on a non-zero exit even when the harness chooses to truncate, would make the difference visible. A regression test would assert that a step exiting non-zero with no output is reported with its code.

Found while replaying gates for #2344; that PR is unaffected — its ten other local gates and the full CI are green.

cyberlife-coder:
Closes #2346.

## Root cause, not the symptom

#2346 recorded the symptom — two lint gates reporting `FAILED` with **no output**, on commands that pass when run by hand — and proposed printing the exit code. That would have made the noise legible without removing it. The cause is in the harness's own `cargo` shim.

`scripts/local-ci.sh` wraps `cargo` in a shell function so that a missing subcommand is classified as a missing tool rather than as a refusal. Inside it:

```sh
proxy=$(command -v "cargo-$sub" 2>/dev/null)
```

`command -v cargo-build` exits **1** for every subcommand cargo implements itself — `build`, `check`, `clean`, `metadata`, `test`. A step whose `run` begins with `set -e` runs the shim under that setting, so the failed assignment ended the shell **there**, before `command cargo` was ever reached. No command ran, nothing was written, and the replayer printed `FAILED` over a blank line.

Measured against the shim's own definition:

```
cargo metadata, with set -e             rc=1  out=[]
cargo metadata, without set -e          rc=0  out=[{"packages":[{"name":"velesdb-core","ver]
cargo fmt (a cargo-fmt shim exists)     rc=0
```

The third line is why only two gates were hit: `cargo fmt` and `cargo clippy` have real `cargo-<sub>` proxies on PATH, so the assignment succeeds. `Drop cached workspace fingerprints` and `Check each velesdb-memory feature in isolation` are the only two lint steps that both enable `set -e` and call a shimless subcommand.

Fix: `|| proxy=""` on that assignment, and the same on `rustup_bin` for a machine without rustup.

## Before and after, on this branch

```
before   ▶ Drop cached workspace fingerprints (force full re-lint) ... FAILED
before   ▶ Check each velesdb-memory feature in isolation ... FAILED
after    ▶ Drop cached workspace fingerprints (force full re-lint) ... ok
after    ▶ Check each velesdb-memory feature in isolation ... ok
```

Full replay on this branch: **41 gates replayed, 0 failed**, 2 tool-missing (`taplo`, `cargo-machete`, absent on this machine), 2 runner-only.

## The report, second

A step can die without writing a line, and `FAILED` over a blank line gives the operator nothing to act on — the mirror image of this script's own rule that a gate it cannot replay is *announced* rather than dropped in silence. The exit code is now printed, and silence is named as silence:

```
  ▶ Spelling (typos) ... FAILED (exit 2)
       error: `raison` should be `reason`, `raisin`
          --> scripts/tests/test_local_ci.py:124:59
```

That is a real red this branch produced and then fixed — a French sentinel in one of the new tests — and it is what the improved report looks like in use. `127` is a missing command, `101` a cargo error, `1` a shell that stopped.

## Tests

Four new tests in `scripts/tests/test_local_ci.py`, already wired at `ci.yml`'s `Test the local gate replayer` step. Two of them were **red against the previous shim with exactly the reported symptom** (`FAILED`, empty output), verified before the fix:

- a strict step calling a shimless subcommand must run it;
- the same on a PATH with no rustup at all;
- a silent non-zero step is reported with its exit code and named as silent;
- a loud one still shows its output, and is *not* described as silent.

`python3 -m unittest scripts.tests.test_local_ci` — 19 passed.


cyberlife-coder:
Rebased on `develop` (now carrying #2350). The local replay on the merged state is the first fully clean one this branch has produced, and it covers both gates that were false reds plus the guard #2350 just added:

```
  ▶ Drop cached workspace fingerprints (force full re-lint) ... ok
  ▶ Check each velesdb-memory feature in isolation ... ok
  ▶ Check every feature that gates code is linted ... ok

Gates replayed: 44 - failed: 0 - tool missing: 2 - runner-only: 2 (of 48 declared)
Every gate this machine could run passes.
```

Before this branch the same replay read `41 replayed - failed: 2`, both with an empty output.

